PR #9, opened 2019-04-02
Prefer https over ssh url for wasi-sysroot submodule

jlb6740:
To avoid some firewall complications https urls are recommend
over ssh:

https://help.github.com/en/articles/which-remote-url-should-i-use

also makes URL type for wasi-sysroot consistent with llvm-project.


sunfishcode:
Makes sense. Thanks for submitting this!

PR #25, opened 2019-04-30
Say "wasm32-wasi" instead of "wasm32-unknown-wasi" in wasi-sdk.

sunfishcode:
Hooray for slightly less clutter.


README.md
```diff
@@ -19,7 +19,7 @@ builds configured to set the default target and sysroot for convenience.
 
 One could also use a standard Clang 8.0, build a sysroot from the sources
```
pchickey: Wait, this won't work with Clang 8.0, but will it work in 8.0.1?

sunfishcode: No, everything here works with Clang 8.0 too.

sunfishcode: If you're watching the LLVM commits, the only patches I landed for this were the ones adding tests for the "wasm32-wasi" triple. The support in the actual code was already present.

pchickey: Ah! I misunderstood that. Thank you for clarifying.

PR #29, opened 2019-05-06
Rename sysroot to wasi-sysroot.

sunfishcode:
This renames /opt/wasi/share/sysroot to /opt/wasi/share/wasi-sysroot.

If the user selects an alternate prefix, this allows wasi-sysroot to
coexist with other packages with less risk of namespace collision. For
example, if the user uses a prefix of /usr/local, this avoids using
/usr/local/share/sysroot, which is fairly generic, and uses
/usr/local/share/wasi-sysroot, which more clearly indicates its purpose.


sunfishcode:
In Lucet, lucet-wasi/build.rs, lucet-wasi-sdk/src/lib.rs, and helpers/install.sh know about the "sysroot" name and will need updating after this.

PR #38, opened 2019-05-31
Disable -fPIC for compiler-rt and libc++abi

yurydelendik:
Since emscripten's -fPIC is supported now in wasm llvm, we need to explicitly disable it.


sunfishcode:
Looks good!

PR #42, opened 2019-07-16
Use ninja generator when using cmake

sbc100:
Ninja has many advantages.  The motivating one is automatic parallelism
according to number of cores on the machine.


sbc100:
Doh! Thanks.

PR #47, opened 2019-07-24
Add Azure Pipelines definition for Linux and macOS

tschneidereit:
This [passes on Linux and macOS](https://dev.azure.com/tschneidereit/WASI%20SDK/_build/results?buildId=88), but has Windows disabled for now, because I pwasn't able to get the build to work there](https://dev.azure.com/tschneidereit/WASI%20SDK/_build/results?buildId=87).

This definition only does the build, but doesn't actually publish it anywhere—we should do that as a follow-up.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,47 @@
+trigger:
+  - master
+
+jobs:
+- job: Build
+  timeoutInMinutes: 360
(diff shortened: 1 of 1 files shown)


sunfishcode:
Cool, this looks like a good place to start with, and we can build from here.

PR #53, opened 2019-09-12
Mention several limitations to be aware of when using wasi-sdk.

sunfishcode:
Mention several limitations to the README. These are all expected to be resolved in the future, but for now, it's useful to let people know what currently works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,10 @@ in the install directory.
 ## Notable Limitations
 
 This repository does not yet support C++ exceptions. C++ code is
-supported only with -fno-exceptions for now. Work on support for
-[exception handling] is underway at the language level.
+supported only with -fno-exceptions for now. Similarly, there is not
+yet support for setjmp/longjmp. Work on support for [exception handling] 
+s underway at the language level which will support both of these
+features.
 
 [exception handling]: https://github.com/WebAssembly/exception-handling/
 
--- a/README.md
+++ b/README.md
@@ -31,3 +31,25 @@ Upstream autoconf now
 For convenience when building packages that aren't yet updated, updated
 config.sub and config.guess files are installed at share/misc/config.\*
 in the install directory.
+
+## Notable Limitations
+
+This repository does not yet support C++ exceptions. C++ code is
+supported only with -fno-exceptions for now. Work on support for
+[exception handling] is underway at the language level.
+
+[exception handling]: https://github.com/WebAssembly/exception-handling/
+
+This repository does not yet support [threads]. Specifically, WASI does
+not yet have an API for creating and managing threads yet, and WASI libc
+does not yet have pthread support.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -31,3 +31,25 @@ Upstream autoconf now
 For convenience when building packages that aren't yet updated, updated
 config.sub and config.guess files are installed at share/misc/config.\*
 in the install directory.
+
+## Notable Limitations
+
+This repository does not yet support C++ exceptions. C++ code is
+supported only with -fno-exceptions for now. Work on support for
+[exception handling] is underway at the language level.
```
kripken: May also want to mention that setjmp/longjmp is not supported yet (and will also work with the exceptions feature).

README.md
```diff
@@ -31,3 +31,25 @@ Upstream autoconf now
...
+This repository does not yet support [threads]. Specifically, WASI does
+not yet have an API for creating and managing threads yet, and WASI libc
+does not yet have pthread support.
+
+[threads]: https://github.com/WebAssembly/threads
+
+This repository does not yet support dynamic libraries. While there are
+[some efforts](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md)
+to design a system for dynamic libraries in wasm, it is still in development
+and not yet generally usable.
```
kripken: Does "in development" refer to the design in the tool-conventions repo? If so I think it's inaccurate, the design is pretty stable and there is a working implementation in production. Or if it refers to the status of the wasi design, or the implementation in this repo, then perhaps that could be clarified.

sunfishcode: This is the wasi-sdk repo; outside of Emscripten and Web contexts, I'm not aware of anyone having even a working prototype yet. Also, to my knowledge the Emscripten implementation doesn't have two-level namespacing yet, so there's more design work to be done.

README.md
sunfishcode: Good idea; I've now added a mention of this.

README.md
kripken: Agreed, yes, but do you not see what I mean about the sentence being ambiguous, though? Sorry if I wasn't clear enough before, let me try again: What I'm suggesting is something like

> This repository does not yet support dynamic libraries. While there is [a design](https://github.com/WebAssembly/tool-conventions/blob/master/DynamicLinking.md) for dynamic libraries in wasm, it is not yet implemented here, and some aspects of the design (two-level namespacing) are still ongoing.

sunfishcode: The linked document introduces itself saying:

> This document describes the current WebAssembly dynamic linking ABI used by emscripten and by the llvm backend when targeting emscripten.
>
> Note: This ABI is still a work in progress. There is no stable ABI yet.

This doesn't imply a finished design outside of Emscripten-specific contexts.

PR #64, opened 2019-10-10
Don't include opt/wasi-sdk paths in tar archive

### LINKED ISSUE #57: Strip "opt/wasi-sdk" from filenames in releases
When we untar a release we end up with e.g.

```
wasi-sdk-6.0/opt/wasi-sdk/bin/clang
```

When `wasi-sdk-6.0/bin/clang` would seem more sensible and concise.

I understand the `opt/wasi-sdk` prefix is useful to using during configuration since it means the SDK will work better when installed at this location.  But when unpacked from tarball I don't think its useful since we can't know that final path.

Looks like a simple change to tar_from_installation.sh

sbc100:
Fixes #57

### DIFF
--- a/tar_from_installation.sh
+++ b/tar_from_installation.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 set -ex
 if [ -n "$1" ]; then
-	export VERSION="$1"
+    VERSION="$1"
 else
-	export VERSION=`./version.sh`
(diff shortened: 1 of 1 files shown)


sbc100:
> My only concern is that this will break environments that automatically install from the tar, but, oh well. This is the right thing to do.

You mean they will need to update their paths?  Yes I think thats OK to do at this point.

PR #68, opened 2019-10-23
Install additional LLVM tools.

sunfishcode:
This installs unqualified `nm`, `ar`, and other symlinks, additional
tools including `llvm-objdump` and `llvm-objcopy`, and additional clang
tools such as `clang-format` and `clang-tidy`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,19 +24,37 @@ build/llvm.BUILT:
 		-DLLVM_DEFAULT_TARGET_TRIPLE=wasm32-wasi \
 		-DLLVM_EXTERNAL_CLANG_SOURCE_DIR=$(LLVM_PROJ_DIR)/clang \
 		-DLLVM_EXTERNAL_LLD_SOURCE_DIR=$(LLVM_PROJ_DIR)/lld \
-		-DLLVM_ENABLE_PROJECTS="lld;clang" \
+		-DLLVM_EXTERNAL_CLANG_TOOLS_EXTRA_SOURCE_DIR=$(LLVM_PROJ_DIR)/clang-tools-extra \
+		-DLLVM_ENABLE_PROJECTS="lld;clang;clang-tools-extra" \
 		-DDEFAULT_SYSROOT=$(PREFIX)/share/wasi-sysroot \
+		-DLLVM_INSTALL_BINUTILS_SYMLINKS=TRUE \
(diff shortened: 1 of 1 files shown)


sunfishcode:
We can do one any time. I'm interested in getting https://github.com/CraneStation/wasi-libc/pull/112 and a few related PRs into a release, but we don't need to block on that if you want to do something sooner.

pchickey:
ok, lets wait for that one and then do a release.

Makefile
```diff
@@ -24,19 +24,37 @@ build/llvm.BUILT:
...
+		install-llvm-objcopy \
+		install-ar \
+		install-ranlib \
+		install-strip \
+		install-nm \
+		install-size \
+		install-objdump \
+		install-objcopy \
+		install-c++filt \
 		llvm-config
```
sbc100: At some point it probably makes sense just install the whole of llvm (not the devel parts but the user parts) rather than trying maintain this long whitelist?

If space saving are what we are after a better approach would probably be to use `LLVM_LINK_LLVM_DYLIB` to avoid duplication between all the binaries?   (This relies on -Wl,-rpath on linux .. so maybe there windows/mac portability issies?, but other investigating).

Makefile
```diff
@@ -24,19 +24,37 @@ build/llvm.BUILT:
...
 		-DDEFAULT_SYSROOT=$(PREFIX)/share/wasi-sysroot \
+		-DLLVM_INSTALL_BINUTILS_SYMLINKS=TRUE \
 		$(LLVM_PROJ_DIR)/llvm
 	ninja -v -C build/llvm \
 		install-clang \
+		install-libclang \
+		install-clang-format \
+		install-clang-tidy \
+		install-clangQuery \
+		install-clang-apply-replacements \
```
sbc100: Do all of these targets existing llvm 8 too?   Do want want to continue to support llvm-8?

sunfishcode: I haven't tested a full build, but they're all resent in the llvm 8 source tree.

Makefile
sunfishcode: If there's a rule to install all the user parts, I'd be happy to switch to it.

And yeah, `LLVM_LINK_LLVM_DYLIB` is nice, but we should investigate its portability first.

PR #70, opened 2019-10-29
Ship wasi-sdk.cmake as part of the SDK

### LINKED ISSUE #56: Ship wasi-sdk.cmake in the SDK bundle?
Perhaps in usr/share/cmake?

Then we could reuse in places like this: https://github.com/WebAssembly/wabt/pull/1068

sbc100:
Fixes #56

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,8 @@ build/libcxxabi.BUILT: build/libcxx.BUILT build/llvm.BUILT
 build/config.BUILT:
 	mkdir -p $(PREFIX)/share/misc
 	cp src/config/config.sub src/config/config.guess $(PREFIX)/share/misc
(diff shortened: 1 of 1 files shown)


sbc100:
For the debian pacakge maybe we could do that, but for the tar archive we don't have any way to install stuff outside of whereever the developer wants to extract the archive.

I guess we could put in in `<archive_root>/share/cmake`?   Would that be preferable?  I'm not sure.  Its not going to be found by cmake automatically either way.

pchickey:
I'd rather it be in `<archive root>/share/cmake` in the tar. I'm happy with changing it to land inside `/usr/share/cmake/` in the deb if that gets it found automatically.

sbc100:
Done

PR #72, opened 2019-10-30
Install strings, don't install libclang, and tidy up install rules

sunfishcode:
Also remove redundant install-* rules. Also, don't install llc, since
it isn't an end-user tool.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,11 +31,8 @@ build/llvm.BUILT:
 		$(LLVM_PROJ_DIR)/llvm
 	ninja -v -C build/llvm \
 		install-clang \
-		install-libclang \
-		install-libclang-headers \
 		install-clang-format \
 		install-clang-tidy \
-		install-clangQuery \
 		install-clang-apply-replacements \
 		install-lld \
 		install-llvm-ranlib \
@@ -50,8 +47,7 @@ build/llvm.BUILT:
 		install-strings \
 		install-objdump \
 		install-objcopy \
-		install-c++filt \
-		install-llvm-config
+		install-c++filt
 	touch build/llvm.BUILT
 
 build/wasi-libc.BUILT: build/llvm.BUILT
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,8 @@ build/llvm.BUILT:
 		install-strings \
 		install-objdump \
 		install-objcopy \
-		install-c++filt
+		install-c++filt \
+		llvm-config
 	touch build/llvm.BUILT
 
 build/wasi-libc.BUILT: build/llvm.BUILT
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -37,25 +37,20 @@ build/llvm.BUILT:
...
 		install-ranlib \
 		install-strip \
 		install-nm \
 		install-size \
+		install-strings \
 		install-objdump \
 		install-objcopy \
 		install-c++filt \
-		llvm-config
+		install-llvm-config
```
sbc100: Is this needed?  I thought llvm-config was used by people wanting to build program against llvm, not with llvm?   Since we are not shipping libs or headers I'm not sure it makes sense to ship this does it?

sunfishcode: We already do install libclang (though not libclang-headers, though I've now fixed that), and while I don't know of anyone using it yet, I do have some potential projects that would want to use it.

sbc100: Don't you think we should ship a minimal version of compiler that useable for building?   Keep it small/simple?   Do you think we have used of this SDK who want to build new llvm-based tools, based on what we are shipping here?  Wouldn't such users already a have general purpose llvm-dev installed, or be building from source?

sunfishcode: Last time I tried a full LLVM install, it installed lots of stuff we don't need.

Would it help if we drop libclang (and headers and llvm-config) for now? Even though I have some ideas in mind for it, I don't urgently need it.

sbc100: I'm happy for you to land this PR in whatever form you see fit.  We can followup with a cleanup of what gets installed later if needed.

sunfishcode: I decided to go ahead and remove libclang and associated targets. They do add about 50M to the install size, and I don't actually have a use for them right now, so if a use does arise, we can re-evaluate.

sbc100: Awesome

Makefile
```diff
@@ -31,31 +31,23 @@ build/llvm.BUILT:
...
 	ninja -v -C build/llvm \
 		install-clang \
-		install-libclang \
 		install-clang-format \
 		install-clang-tidy \
-		install-clangQuery \
 		install-clang-apply-replacements \
 		install-lld \
-		install-llc \
-		install-llvm-ar \
```
sbc100: For ranlib and strip we have `install-llvm-ranlib` in addition to `install-ranlib`.  Do they do different things?  If you are going to remove the `install-llvm-` variants presumably you should do so for ranlib and strip too?

sunfishcode: They do different things. I realize this is non-obvious, and it would be way nicer if LLVM just gave us a single install rule instead of doing all this manually. I don't know of such a rule.

sbc100: I found some llvm methods for doing what we want (one of them is called LLVM_INSTALL_TOOLCHAIN_ONLY) but we can do that in a followup.  Happy for this to land as is.

sunfishcode:
Yes, thanks!

PR #75, opened 2019-10-30
Remove redundant llvm cmake args

sbc100:


sbc100:
yeah, those a legacy settings from the pre-monorepo world I believe.

PR #76, opened 2019-10-31
Don't try to strip llc (it was removed in 5ae9322)

sbc100:


sbc100:
Looks like this passed CI because we only test `build` and not `package` as part of CI.

PR #80, opened 2019-11-08
switch to Github Actions, upload built artifacts to tags

pchickey:


pchickey:
Anyone know what to make of this windows failure? https://github.com/CraneStation/wasi-sdk/pull/80/checks?check_run_id=293676924#step:7:3535

sunfishcode:
The error might be a bug: https://reviews.llvm.org/D64077#1731233 (see the linked comment).

.github/actions/github-release/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "wasmtime-github-release",
```
sbc100: Is this is an actual close on wasmtime-github-releases can we not reference it as a `@dependency` rather than cloning it?   Or will it need to be locally modified?

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
+name: CI
+on:
+  push:
+    branches: [master]
+    tags-ignore: [dev]
```
sbc100: We don't have a dev branch here (that I know of anyway).

pchickey:
Hey, thanks for the reminder to get back to this. I'll try to get it working this week.

Will delete the azure pipelines integration once github actions are working properly.

.github/actions/github-release/package.json
pchickey: I think the idea behind vendoring it in as source code was so we're not trusting external code with release assets. Additionally, we can remove a bunch of the logic around the dev branch, I think. cc @alexcrichton

alexcrichton: Ideally this would use things like https://github.com/actions/create-release or https://github.com/actions/upload-artifact but those didn't quite work for wasmtime's purposes.

In terms of vendoring and such it's probably easier to copy in whatever's needed here and have it evolve as this repo evolves, once you get this working it tends to not change too much anyway.

sbc100: tOK I see.  Seeing as we don't have wasmtimes requirements perhaps we can just use the off-the-shelf stuff.   Much nicer that having some bespoke JS thing (IMHO).

Per my previous suggestion how about removing the artifact uploading stuff, and have this PR simply switch from azure to github actions, then we can experiment and bikeshed the artifact upload stuff in a followup?

pchickey: Yeah, that is fine. I'll re-scope this PR to do only that.

PR #82, opened 2019-11-25
Switch to Github Actions CI

pchickey:
Creates a Github Actions CI configuration, which should do the exact same thing as the Azure Pipelines CI did.

Deletes Azure Pipelines configuration.

Replaces #80


pchickey:
It appears to spin up 5 separate build jobs for each platform, when it should only do 2 - one for push, one for PR (which tests the merge commit iirc). I closed the last PR to see if some leftover state was causing that weirdness.

sbc100:
Yeah, I thinks the expected behavior (its what travis does too I think).   One is testing the actual thing you pushed, one is testing the PR (i.e. after integration with master).   Shame its not smart enough to know that that is the same thing in almost all cases when you are working on top of master anyway.

sbc100:
Can can probably make do with `pull_requests` to be tested, and `push` only for master?

pchickey:
right, i expect it to test those two, on mac os and on linux, so 4 jobs total. But on my screen there are 10 jobs total, and the other 6 are all dupes.

The good news is that #83, which has an identical workflow, seems to have the right number of jobs.

sbc100:
Oh sorry, ignore me.. that doe s indeed look strange.

PR #83, opened 2019-11-25
Github Actions uploads built artifacts

pchickey:
This changes the build scripts to put the artifacts we use for releases in `dist/`, and has Github Actions upload those artifacts to Github.

I tried doing some yaml programming to automate putting those artifacts into a release, but I wasn't having a great time of it, but this should make it much easier to produce releases by manually downloading artifacts and then attaching them to a release, rather than build them locally.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,5 +42,6 @@ jobs:
         name: Build
       - uses: actions/upload-artifact@v1
         with:
-          name: ${{ format( 'packages-{0}.tgz', matrix.os) }}
+          # Upload the dist folder. Give it a name according to the OS it was built for.
+          name: ${{ format( 'dist-{0}.tgz', matrix.os) }}
           path: dist
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,3 +40,7 @@ jobs:
       - run: sudo make -j4 package
         shell: bash
         name: Build
+      - uses: actions/upload-artifact@v1
+        with:
+          name: ${{ format( 'packages-{0}.tgz', matrix.os) }}
+          path: dist
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,7 +14,7 @@ jobs:
           # llvm 9 build is broken on windows for now. Re-enable this when it is fixed:
           # - windows-latest
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v1
         with:
           submodules: true
       - run: |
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -149,11 +149,11 @@ build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libc
...
+package: dist
 
-build/package.BUILT: build strip
-	command -v dpkg-deb >/dev/null && ./deb_from_installation.sh || true
-	./tar_from_installation.sh
-	touch build/package.BUILT
+dist: build strip
+	mkdir -p dist
+	command -v dpkg-deb >/dev/null && ./deb_from_installation.sh $(shell pwd)/dist || true
+	./tar_from_installation.sh $(shell pwd)/dist
```
sbc100: Why not continue to use stamp files for consistency?  i.e. can you move into the "dist" directory with out changing the way `build/package.BUILT` is used.   Having `package` depend a `dist` seems add because `dist` in this case is a directory rather than a `.phony` targets.

deb_from_installation.sh
```diff
@@ -1,9 +1,15 @@
 #!/usr/bin/env sh
 set -ex
 if [ -n "$1" ]; then
-	export VERSION="$1"
+    OUTDIR=$1
 else
-	export VERSION=`./version.sh`
+    OUTDIR=$PWD/build
```
sbc100: Default to "$PWD/dist"?

.github/workflows/main.yml
```diff
@@ -40,3 +40,7 @@ jobs:
       - run: sudo make -j4 package
         shell: bash
         name: Build
+      - uses: actions/upload-artifact@v1
+        with:
+          name: ${{ format( 'packages-{0}.tgz', matrix.os) }}
```
sbc100: Can you explain this line?

What is `packages-`?   

Is there any way to test this?    Don't we want to limit this to only uploading for tags?

Makefile
pchickey: Fixed, thanks

.github/workflows/main.yml
pchickey: I renamed it `dist` to be more consistent, and added a comment. 

The upload appears in the "Actions" tab on GH, and its not actually a release. It just allows us to download the Actions-built contents of `dist`, untar that into the actual built artifacts, and then upload those to a tag. I gave up on programming that sequence in yml, so we can do that by hand until someone dles.

deb_from_installation.sh
pchickey: Fixed

.github/workflows/main.yml
sbc100: Sweet!  So we can get a copy of the sdk for each pull request.  Nice.  Out of interest do you know are disk space limits enforced on this kind of resource?

PR #88, opened 2019-12-20
Add a rudimentary testsuite.

sunfishcode:
This adds a very primitive test harness and tests relevant to the recent
changes to how program startup works, as well as the upcoming changes
to support LTO.


tests/testcase.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
sbc100: This is going to outgrow the (my) accepted level of shell script usage very quickly.

Would you object to me re-writing this in python at some point?

tests/run.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -ueo pipefail
+
```
sbc100: Comment/docs here?

tests/empty-expected.wat
```diff
@@ -0,0 +1,13 @@
+(module
```
sbc100: Is this file used anywhere?  I can't seem to see any references.

sunfishcode: Nope, removed!

tests/run.sh
sunfishcode: Added

tests/testcase.sh
```diff
@@ -0,0 +1,57 @@
...
+echo "Testing $input..."
+"$clang" $options $file_options "$input" -o "$wasm"
+if [ -e "$input.stdin" ]; then
+  stdin="$input.stdin"
+else
+  stdin="/dev/null"
+fi
+
+exit_status=0
+"$runwasm" "$wasm" \
```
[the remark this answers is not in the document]
sunfishcode: Renamed to runwasi.

tests/testcase.sh
```diff
@@ -0,0 +1,57 @@
...
+else
+  stderr_expected="/dev/null"
+fi
+if [ -e "$input.exit_status.expected" ]; then
+  exit_status_expected="$input.exit_status.expected"
+else
+  exit_status_expected=../exit_status_zero
+fi
+
+diff -u "$stderr_expected" "$stderr_observed"
```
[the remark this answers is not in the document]
sunfishcode: Comments added.

tests/testcase.sh
sunfishcode: No objections. I find what these scripts are doing simpler to write in bash than python, but I'm ok with them being written in python.

sbc100:
Maybe add a `make check` target that runs `tests/run.sh` with the just-built SDK?  And then add `make check` to `.github/workflows/main.yml`?    Or we can do that in a followup if you prefer?

sunfishcode:
> One general comment: Can we run the tests (or some subset) in compile-only mode? That way somebody could run some basic tests without a wasi runtime?
> 
> That would also allow us to add this to CI right away without worrying about that existence of the wasi runtime in the CI (which could be a followup).

I like that idea. I've now modified the script to allow it to be run without a runtime, and it still compiles everything but just disables running and the subsequent comparisons.

> Also, apologies if these changes are not ready for review yet. I'm not sure what your model is for that? Would you rather I wait until you nominate reviewers?

I should probably get in the habit of nominating reviewers for these repos. In any case, anyone is welcome to post review comments, nominated reviewer or not :-).

sunfishcode:
> Maybe add a `make check` target that runs `tests/run.sh` with the just-built SDK? And then add `make check` to `.github/workflows/main.yml`? Or we can do that in a followup if you prefer?

I've now added a patch to this PR to do this.

PR #89, opened 2019-12-20
wasi-sdk support for full LTO.

sunfishcode:
Add support for building libcxx and libcxxabi in LTO mode, along aside
the regular non-LTO builds.

Pass CLANG_VERSION to the wasi-libc build so that it can build itself in
LTO and non-LTO modes.

And drop the explicit -O1 from the compiler-rt build, and just go with
the default optimization settings.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ clean:
 
 build/llvm.BUILT:
 	mkdir -p build/llvm
-	cd build/llvm; cmake -G Ninja \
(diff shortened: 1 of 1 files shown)


sunfishcode:
This is now rebased on top of #106, so the changes are much clearer now.

abrown:
@sunfishcode, this seems like a nice feature to have, do you want to rebase it?

PR #92, opened 2020-01-06
Add NINJA_FLAGS support

TerrorJack:
`ninja` lacks support for using an environment variable to supply additional flags. Since `ninja` is invoked by `make` in this project, this PR adds a `NINJA_FLAGS` make variable, to allow additional flags to be passed to `ninja` when running `make` (e.g. explicit `-jN` when auto-detected CPU core count is unreliable on CI). When `NINJA_FLAGS` is unspecified, the default build behavior is not affected.


sbc100:
Nice.   Maybe we should remove the `-v` default too since it can easily be re-added via NINJA_FLAGS?

pchickey:
I'm fine with removing it, but it doesn't bother me to have it enabled by default.

PR #94, opened 2020-01-13
use compnerd/cmake's build of wasi-libc

pchickey:
Using this PR to get the CI system to try building with wasi-libc's new CMamke based build system, by @compnerd

https://github.com/CraneStation/wasi-libc/pull/154#issuecomment-573383839

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,8 +60,9 @@ build/wasi-libc.BUILT: build/llvm.BUILT
 		-DCMAKE_C_COMPILER=$(PREFIX)/bin/clang \
 		-DCMAKE_C_COMPILER_WORKS=YES \
 		$(ROOT_DIR)/src/wasi-libc
-	ninja $(NINJA_FLAGS) -v -C build/wasi-libc \
-		install
+	ninja $(NINJA_FLAGS) -v -C build/wasi-libc
+	mkdir -p $(PREFIX)/share
+	cp -R build/wasi-libc/sysroot $(PREFIX)/share/wasi-sysroot
 	touch build/wasi-libc.BUILT
 
 build/compiler-rt.BUILT: build/wasi-libc.BUILT build/llvm.BUILT
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,6 @@ build/llvm.BUILT:
 
 build/wasi-libc.BUILT: build/llvm.BUILT
 	cmake -G Ninja \
-		--install \
 		-DCMAKE_SYSTEM_NAME=Generic \
 		-DCMAKE_AR=$(PREFIX)/bin/ar \
 		-DCMAKE_NM=$(PREFIX)/bin/nm \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -52,12 +52,17 @@ build/llvm.BUILT:
...
 
 build/wasi-libc.BUILT: build/llvm.BUILT
-	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
-		WASM_CC=$(PREFIX)/bin/clang \
-		SYSROOT=$(PREFIX)/share/wasi-sysroot
+	cmake -G Ninja \
+		--install \
+		-DCMAKE_SYSTEM_NAME=Generic \
+		-DCMAKE_AR=$(ROOT_DIR)/bin/llvm-ar \
+		-DCMAKE_NM=$(ROOT_DIR)/bin/llvm-nm \
```
sunfishcode: This is pre-existing, but we can remove the "llvm-" from these commands, as we install a plain "nm" and "ar" now.

compnerd:
Drop the explicit `install` target and I think that it should work.  I didn't really see an install target previously, so I did not create one.  Currently, the CMake build will generate a complete sysroot under the `sysroot` directory in the build tree.

pchickey:
Subsequent build steps (compiler-rt) are calling the clang built prior to build, and requires some headers provided by wasi-libc. Clang's system include path is searching at $(PREFIX)/share/wasi-sysroot. So, we need to install wasi-libc to that location for the rest of the build to work, and for it to end up in the packages created by this repo.

compnerd:
@pchickey sorry if what I was saying was unclear.  What I meant is that something like this should work:

```
cmake -G Ninja \
 		-DCMAKE_SYSTEM_NAME=Generic \
		 -DCMAKE_SYSTEM_NAME=Generic \
 		-DCMAKE_AR=$(PREFIX)/bin/ar \
		 -DCMAKE_AR=$(PREFIX)/bin/ar \
 		-DCMAKE_NM=$(PREFIX)/bin/nm \
		 -DCMAKE_NM=$(PREFIX)/bin/nm \
 		-DCMAKE_C_COMPILER=$(PREFIX)/bin/clang \
		 -DCMAKE_C_COMPILER=$(PREFIX)/bin/clang \
 		-DCMAKE_C_COMPILER_WORKS=YES \
		 -DCMAKE_C_COMPILER_WORKS=YES \
		 -B build/wasi-libc \
 		-S $(ROOT_DIR)/src/wasi-libc
cmake --build build/wasi-libc
```

That will generate a proper sysroot in `build/wasi-libc/sysroot`.  A `cp -r build/wasi-libc/sysroot $(PREFIX)/share/wasi-sysroot` should be sufficient to "install"

pchickey:
@compnerd thanks, that fix got us through compiler-rt, but now libcxx is failing. Any ideas?

compnerd:
```
-- Performing Test LIBCXX_SUPPORTS_STD_EQ_CXX14_FLAG
-- Performing Test LIBCXX_SUPPORTS_STD_EQ_CXX14_FLAG - Failed
-- Performing Test LIBCXX_SUPPORTS_STD_COLON_CXX14_FLAG
-- Performing Test LIBCXX_SUPPORTS_STD_COLON_CXX14_FLAG - Failed
  C++11 or greater is required but the compiler does not support c++14
```
Seems that `-std=c++14` and `/std:c++14` are both failing.  Unfortunately, I cannot tell what is failing in the test :(.  You *could* theoretically try `-D LIBCXX_SUPPORTS_STD_EQ_CXX14_FLAG=YES` as a workaround.

pchickey:
I never got back to this and it isn't a priority, so closing for now. If anyone else wants to take this work on, feel free to re-open.

PR #97, opened 2020-01-17
Add several tests to test printf of various configurations.

sunfishcode:
wasi-libc builds special versions of printf and other routines optimized
for the case where long double and floating-point in general are not
needed. Ensure that everything still compiles and runs.

### DIFF
--- a/tests/general/printf-no-long-double.c
+++ b/tests/general/printf-no-long-double.c
@@ -1,0 +1,8 @@
+#include <stdio.h>
+
+volatile double x = 42.0;
+
+int main(void) {
+    printf("the answer is %f\n", x);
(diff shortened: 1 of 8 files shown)


tests/general/printf-no-long-double.c
```diff
@@ -0,0 +1,8 @@
+#include <stdio.h>
+
+volatile double x = 42.0;
```
sbc100: Is volatile needed here?

sunfishcode: Just an abundance of caution; the point is to test the `libc` printf; if a compiler ever somehow constant-folds this into the printf string, that would defeat the point.

PR #100, opened 2020-02-19
Add a LICENSE file, "Apache-2.0 WITH LLVM-exception".

sunfishcode:
This uses Apache-2.0, which is the same as what's used in other
WebAssembly CG repositories, plus the LLVM-exception which provides for
GPLv2 compatibility and makes it easier to move code between libraries
and compilers.

@tschneidereit


tschneidereit:
LGTM, thanks!

PR #104, opened 2020-02-21
Various github action cleanups / improvements

sbc100:
- Don't build pushes to non-main branches, only build the PR itself.
- Remove special windows llvm-nm installation
  The SDK build process itself builds llvm so we don't need to download
  it.
- Make the `name` the first item in each step.  I peronally find
  this more readable.


sbc100:
The windows build is a lot slower.. (good things its broken :)

PR #105, opened 2020-03-11
Add tests for assert, abort, and environ.

sunfishcode:
To support this, add a mechanism for filtering test output to filter
out uninteresting diffs.

This adds some coverage for the code being changed in
https://github.com/WebAssembly/wasi-libc/pull/184.


tests/general/assert-fail.c.stderr.expected
```diff
@@ -0,0 +1,7 @@
+Assertion failed: false (assert-fail.c: main: 5)
+Error: failed to run main module `assert-fail.c.---.wasm`
+
+Caused by:
+    0: failed to invoke `_start`
+    1: wasm trap: unreachable, source location: @----
+       wasm backtrace:
```
sbc100: Where is this output coming from?  Is it coming from some wasmtime-internal place?

sunfishcode: Yeah. If other engines have a different format, we can adjust the filter script as needed.

PR #108, opened 2020-03-24
Update LLVM to 10.x

abrown:


Makefile
```diff
@@ -164,7 +164,7 @@ build/config.BUILT:
 build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libcxxabi.BUILT build/libcxx.BUILT build/config.BUILT
 
 strip: build/llvm.BUILT
-	cd $(PREFIX)/bin; strip clang-9 lld llvm-ar
```
sbc100: I would imagine that all the clang binaries are symbolic links to each other?  I wonder why this wasn't just "clang" from the very beginning?

sbc100:
Hmm.. did llvm 10 actually get released?  I guess I missed the announcement.

sbc100:
Wait.. http://lists.llvm.org/pipermail/llvm-dev/2020-March/140233.html .. that message was sent after this PR..  talk about fast upgrades !

abrown:
> Wait.. http://lists.llvm.org/pipermail/llvm-dev/2020-March/140233.html .. that message was sent after this PR.. talk about fast upgrades !

Ha! I wasn't even aware... I actually wanted more SIMD functionality and @sunfishcode suggested the `release/10.x` branch.

Makefile
abrown: Should it actually be something like `find $(PREFIX)/bin -type f -executable -exec strip {} \;` to get all of the binaries?

sbc100: yes, something like that sounds better.

PR #109, opened 2020-03-24
Strip symbols of all binaries

abrown:
@sbc100, this resolves the discussion we had in #108 about stripping symbols from all binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -164,7 +164,7 @@ build/config.BUILT:
 build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libcxxabi.BUILT build/libcxx.BUILT build/config.BUILT
 
 strip: build/llvm.BUILT
-	cd $(PREFIX)/bin; strip clang-* clang++-* lld llvm-ar
+	./strip_symbols.sh $(PREFIX)
 
 package: build/package.BUILT
 
--- a/strip_symbols.sh
+++ b/strip_symbols.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env sh
+set -e
+
+DIRECTORY=${1:-/opt/wasi-sdk/bin}
+EXECUTABLES=$(find ${DIRECTORY} -type f -executable)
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -164,7 +164,7 @@ build/config.BUILT:
 build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libcxxabi.BUILT build/libcxx.BUILT build/config.BUILT
 
 strip: build/llvm.BUILT
-	cd $(PREFIX)/bin; strip clang-* clang++-* lld llvm-ar
+	./strip_symbols.sh $(PREFIX)/bin
```
sbc100: Given the existence of this script I wonder if we should just pass `$(PREFIX)`?   Maybe there are ELF files in other directories and we want to strip those too, no?

abrown: I'm not opposed to doing that; I just was concerned about stripping symbols where I shouldn't. Are we sure it's ok to do it on the entire `$(PREFIX)`?

sbc100: This is the approach the debian packages normally take IIRC.. strip *all* ELF files by default.  I see no reason not to do it here... but I could be missing something.

sunfishcode:
Testing for "ELF" is too narrow since we also support on-ELF hosts such as macOS. Would it work to just unconditionally run strip on every file under `bin` and ignore errors?

Alternatively, if we could add "-s" to the link command-line, that would cause it to strip executables too.

abrown:
@sunfishcode, I didn't really know all the places `-s` might be needed so I tried the "strip everything and ignore errors" approach. Note that due to the discussion above with @sbc100 (resolved now) we are stripping everything, not just the `bin` directory.

sbc100:
I think post-stripping is probably best. 

The current approach of this patch is to strip anything that `find` things is executable.  Seems reasonable to me that we just stick with that approach. Dan?

sunfishcode:
Sounds reasonable.

PR #113, opened 2020-04-02
latest wasi-libc (NFC)

pchickey:
Readme changes, and the latest wasi repo
 https://github.com/WebAssembly/wasi-libc/pull/191


pchickey:
I'm not sure where I picked up the habit but it seems to be standard at my current gig and the prior one. I'm also surprised theres not tool support for it.

PR #116, opened 2020-04-06
Fix `strip_symbols.sh` on mac; bump libc to include utime.h

pchickey:
Fix for `strip_symbols.sh` should fix Mac OS CI.

Bump of libc submodule gets latest change to have utime.h included again.

These are the last blocking issues for the wasi-sdk-9 release, so once we merge this I'll tag it.


pchickey:
Pushed the tag and building release now

PR #123, opened 2020-04-11
Build against ubuntu xenial

### LINKED ISSUE #98: wasi-sdk 8.0 libstdc++ required version is too high
On an Ubuntu Xenial system, I see:
```
/opt/wasi-sdk/bin/clang: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.22\' not found (required by /opt/wasi-sdk/bin/clang)
```
This hasn't been an issue with past releases, and seems to be new since basing the Dockerfile on Debian stretch rather than Xenial. Could we consider switching back to Xenial, or an older version of Debian that would pull in a compatible `libstdc++`?

sbc100:
Fixes #98

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 # Use a relatively old/stable distro here to maximize the supported platforms
 # and avoid depending on more recent version of, say, libc.
-FROM debian:stretch
+# Here we choose Xenial 16.04 which mean we also support Debian from stretch
+# (releases 2017) onwards.
+FROM ubuntu:xenial
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


pchickey:
Can you please fixup the name of the upload artifact as well? https://github.com/WebAssembly/wasi-sdk/blob/master/.github/workflows/main.yml#L67 I named it `dist-debian-stretch` to differentiate it from `dist-ubuntu-latest` and `dist-macos-latest`. I guess now it should be called `dist-ubuntu-xenial`?

sbc100:
Oops sure.

PR #134, opened 2020-05-17
Updated expected results

sbc100:
These were generated using latest wasmtime release (0.16.0) and the
wasi-sdk 10 release binaries.

### DIFF
--- a/tests/general/abort.c.stderr.expected
+++ b/tests/general/abort.c.stderr.expected
@@ -2,5 +2,5 @@ Error: failed to run main module `abort.c.---.wasm`
 
 Caused by:
     0: failed to invoke `_start`
-    1: wasm trap: unreachable, source location: @----
+    1: wasm trap: unreachable
        wasm backtrace:
--- a/tests/general/assert-fail.c.stderr.expected
+++ b/tests/general/assert-fail.c.stderr.expected
@@ -3,5 +3,5 @@ Error: failed to run main module `assert-fail.c.---.wasm`
 
 Caused by:
     0: failed to invoke `_start`
-    1: wasm trap: unreachable, source location: @----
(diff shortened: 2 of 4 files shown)


sbc100:
Are these differences to be expected @sunfishcode ?

sbc100:
I'm also seeing a failure in the stat.c tests (seems like lstat is failing):

```
--- /dev/null	2020-04-08 17:47:21.816001886 -0700
+++ stat.c.-O0.stderr.observed	2020-05-16 18:42:23.772233121 -0700
@@ -0,0 +1,15 @@
+Assertion failed: n == 0 (stat.c: main: 59)
+Error: failed to run main module `stat.c.-O0.wasm`
+
+Caused by:
+    0: failed to invoke `_start`
+    1: wasm trap: unreachable
+       wasm backtrace:
+         0: 0x3198 - <unknown>!abort
+         1: 0x3ae6 - <unknown>!__assert_fail
+         2:  0xba2 - <unknown>!main
+         3: 0x399a - <unknown>!__main_argc_argv
+         4: 0x3a16 - <unknown>!__main_void
+         5: 0x3a7e - <unknown>!__original_main
+         6:  0x2e6 - <unknown>!_start
+  
 ```

sunfishcode:
For the trap messages, Wasmtime recently changed its output to format the backtrace offsets differently. The change you have here is fine for now, though of course it'd be good to find a more general way of handling this too.

For the destructor changes, the change here is incorrect. The difference is due to a bug that [this patch](https://reviews.llvm.org/D70685) fixes, and that fix isn't in the 10.x release branch. We probably need to disable that test when running with clang <= 10.

The stat.c failure appears to be a recent regression in Wasmtime. We're investigating.

Issue #133, opened 2020-05-14
Autotools project fails to compile with -save-temps


orestisfl:
Hello, I am trying to compile [libsodium](https://github.com/jedisct1/libsodium) which has a [shell script](https://github.com/jedisct1/libsodium/blob/master/dist-build/wasm32-wasi.sh) to build with WebAssembly support, using wasi-sdk.

The script executes correctly by default but when I try to add `-save-temps` to the CFLAGS, compilation stops after a bit with the following error:

```
  CC       crypto_onetimeauth/poly1305/libsodium_la-onetimeauth_poly1305.lo
  CC       crypto_onetimeauth/poly1305/donna/libsodium_la-poly1305_donna.lo
  CC       crypto_pwhash/argon2/libsodium_la-argon2-core.lo
  CC       crypto_pwhash/argon2/libsodium_la-argon2-encoding.lo
  CC       crypto_pwhash/argon2/libsodium_la-argon2-fill-block-ref.lo
  CC       crypto_pwhash/argon2/libsodium_la-argon2.lo
  CC       crypto_pwhash/argon2/libsodium_la-blake2b-long.lo
onetimeauth_poly1305.s:15:20: error: Unexpected token in operand: (
        i32.call_indirect       (i32, i32, i64, i32) -> (i32)
                                ^
onetimeauth_poly1305.s:33:20: error: Unexpected token in operand: (
        i32.call_indirect       (i32, i32, i64, i32) -> (i32)
                                ^
onetimeauth_poly1305.s:49:20: error: Unexpected token in operand: (
        i32.call_indirect       (i32, i32) -> (i32)
                                ^
onetimeauth_poly1305.s:66:20: error: Unexpected token in operand: (
        i32.call_indirect       (i32, i32, i64) -> (i32)
                                ^
onetimeauth_poly1305.s:82:20: error: Unexpected token in operand: (
        i32.call_indirect       (i32, i32) -> (i32)
```

I have confirmed that if I remove wasi-sdk and compile without a wasm target & modified sysroot, compilation works with `-save-temps` and I can see the temporary files as a result.

To illustrate, I have created 2 docker files for easy reproduction: https://gist.github.com/orestisfl/5b5b333d5da7001dc799abea719ad9cf
One patches out wasi-sdk from the script and adds `-save-temps` and one that just adds the `-save-temps` flag to `$CFLAGS`.

```shell
[...]

PR #136, opened 2020-05-26
Remove branch from LLVM submodule

orestisfl:
The update to LLVM 10.x was done in #108 but the submodule branch was not updated

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,6 @@
 [submodule "src/llvm-project"]
 	path = src/llvm-project
 	url = https://github.com/llvm/llvm-project
-	branch = release/9.x
 [submodule "src/wasi-libc"]
 	path = src/wasi-libc
 	url = https://github.com/CraneStation/wasi-libc


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "src/llvm-project"]
 	path = src/llvm-project
 	url = https://github.com/llvm/llvm-project
-	branch = release/9.x
```
sbc100, line 4: Does this matter?  Can we just remove this line instead?  Isn't the sub module pinned to a certain git commit, regardless of the branch?    (Sorry I'm not a big user of submodule so I'm probably missing something here).

pchickey: This only ends up mattering for the behavior of `git pull` in a submodule, I believe? For the way we end up using the submodule, I don't think it matters much one way or the other. We can just leave it at whatever git set it to.

sbc100: I think that case maybe lets just drop this line, since otherwise the same thing will happen every 6 months when we forget to update it.  Also I imagine the updating between branches will more common than updating with a branch.. which lessens that value of having this line I think?

pchickey: Agreed then, lets drop the line.

orestisfl: This matters if you want to run `git submodule update --remote`. As it is now, the update will set the branch "backwards" to release/9.x. It also makes the selection of the target more transparent. Right now it is at least a bit misleading. I was bitten by it when I tried to update LLVM as mentioned in https://github.com/WebAssembly/wasi-sdk/issues/133.

I can re-purpose the PR to delete the line if you prefer it though.

sbc100: Yes, I think deleting the misleading line would be better.

orestisfl: up

PR #137, opened 2020-05-26
Add a basic mmap and munmap testcase.

sunfishcode:
This adds a basic testcase for `mmap` and `munmap`, which covers the fixes reported in https://github.com/WebAssembly/wasi-libc/pull/197 and https://github.com/WebAssembly/wasi-libc/pull/198.

### DIFF
--- a/tests/general/mmap.c
+++ b/tests/general/mmap.c
@@ -16,8 +16,8 @@
     return EXIT_FAILURE;         \
   } while (0)
 
-#define OFFSET "10726"
-#define LENGTH "143"
+#define OFFSET 10726
+#define LENGTH 143
 
 int main(int argc, char *argv[]) {
   if (argc != 2) {
@@ -38,7 +38,7 @@ int main(int argc, char *argv[]) {
   if (fstat(fd, &stat_buf) != 0)
     perror_and_exit("fstat");
 
-  off_t offset = atoll(OFFSET);
+  off_t offset = OFFSET;
   if (offset < 0) {
     fprintf(stderr, "negative offset\n");
     return EXIT_FAILURE;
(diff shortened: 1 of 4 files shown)


tests/general/mmap.c
```diff
@@ -0,0 +1,87 @@
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <limits.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#ifdef __GLIBC__
```
sbc100, line 8: Is there a requirement that these tests build outside of wasi-sdk too?  Do we have a test runner mode that does that?

tests/general/mmap.c
```diff
@@ -0,0 +1,87 @@
...
+#include <unistd.h>
+
+#define perror_and_exit(message) \
+  do {                           \
+    perror(message);             \
+    return EXIT_FAILURE;         \
+  } while (0)
+
+#define OFFSET "10726"
+#define LENGTH "143"
```
sbc100, line 20: Why parse these strings?  Why not just `static const size_t LENGTH` and `static const off_t OFFSET`?  Maybe they were originally argv string?  Seems like this test code would be a lot more precise and concise without the parsing and validation.

tests/general/mmap.c
sunfishcode: We don't have that right now, though we may want that in the future.

tests/general/mmap.c
sunfishcode: Done. Yeah, these are essentially parameters. I've now converted them to integers to avoid the atoll/strtoll.

tests/general/mmap.c
```diff
@@ -0,0 +1,87 @@
...
+
+  off_t offset = OFFSET;
+  if (offset < 0) {
+    fprintf(stderr, "negative offset\n");
+    return EXIT_FAILURE;
+  }
+  if (offset > (off_t)SIZE_MAX) {
+    fprintf(stderr, "offset overflow\n");
+    return EXIT_FAILURE;
+  }
```
sbc100, line 49: Why not drop these checks too, since the value is constant?    lgtm either way

sunfishcode: The offset value is effectively a parameter to the test, so it's nice to sanity-check it.

PR #140, opened 2020-06-01
Test signals

sunfishcode:
This adds some basic tests for signal.h functionality added in https://github.com/WebAssembly/wasi-libc/pull/183


tests/general/sigabrt.c
```diff
@@ -0,0 +1,11 @@
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+int main(void)
+{
```
sbc100, line 6: The style here seems to be curlys on the same line as the declaration?

tests/general/signals.c
```diff
@@ -0,0 +1,88 @@
...
+#endif
+#ifndef SIGINT
+#error SIGINT not defined
+#endif
+#ifndef SIGSEGV
+#error SIGSEGV not defined
+#endif
+#ifndef SIGTERM
+#error SIGTERM not defined
+#endif
```
sbc100, line 26: Are these really necessary?    Seems like just using them should be fine.

sunfishcode: I've now extended the test to actually use all these.

tests/general/sigabrt.c
sunfishcode: Fixed.

sunfishcode:
The CI failure is due to needing a newer wasi-libc; cc #141.

sunfishcode:
CI passed.

pchickey:
the built artifacts from here pass lucet's test suite as well, so we're set to go. I'll pushed the tag, we'll let CI build us the release artifacts now.

PR #143, opened 2020-06-14
Build and test SDK without requiring root (sudo) access

sbc100:
We still build with the `/opt/wasi-sdk` prefix but we don't actually
install into that directory.  Instead we install into `build/install`.

The main different is that this requires the `--sysroot` flag when
running clang to build the testsuite.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ ESCAPE_SLASH=/
 # assuming we're running under msys2 (git-bash), PATH needs /c/foo format directories (because
 # it itself is :-delimited)
 DESTDIR=$(abspath build/install)
-BUILD_PREFIX=$(DESTDIR)$(shell cygpath.exe -u $(PREFIX))
+BUILD_PREFIX=$(DESTDIR)/wasi-sdk
 
 else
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,10 +7,10 @@ LLVM_PROJ_DIR?=$(ROOT_DIR)/src/llvm-project
 # Windows needs munging
 ifeq ($(OS),Windows_NT)
 
-PREFIX?=c:/wasi-sdk
+PREFIX=c:/wasi-sdk
```
pchickey, line 10: I'd prefer to leave the `?` in here so these settings can be overridden downstream

Makefile
```diff
@@ -22,14 +22,18 @@ ESCAPE_SLASH=/
...
-PATH_PREFIX=$(shell cygpath.exe -u $(PREFIX))
+DESTDIR=$(abspath build/install)
+BUILD_PREFIX=$(DESTDIR)/wasi-sdk
 
-endif
+else
 
-PREFIX?=/opt/wasi-sdk
-PATH_PREFIX?=$(PREFIX)
+PREFIX=/opt/wasi-sdk
```
pchickey, line 30: Same here, I configure this prefix in a downstream package for an internal deb build, so I'd prefer to keep it with `?=`

PR #150, opened 2020-08-04
Remove `-Wl,--no-threads` from linker flags

sbc100:
I'm not sure why this was in here as it pre-dates the initial
commit.  I imagine it could be that there were bugs in the early version
of wasm-ld related to threading.

In recent version of lld this flag has been renamed:
  https://reviews.llvm.org/D76885


wasi-sdk.cmake
```diff
@@ -22,7 +22,6 @@ set(CMAKE_C_COMPILER_TARGET ${triple} CACHE STRING "wasi-sdk build")
 set(CMAKE_CXX_COMPILER_TARGET ${triple} CACHE STRING "wasi-sdk build")
 set(CMAKE_C_FLAGS "-v" CACHE STRING "wasi-sdk build")
 set(CMAKE_CXX_FLAGS "-v -std=c++11" CACHE STRING "wasi-sdk build")
```
sbc100, line 24: Should we remove these C flags as well?  Wouldn't it be best to leave it up individual projects to decide which C++ version they want.. .and doesn't `-v` generate a ton of verbose output?

sbc100:
Closing in favor or #151

PR #151, opened 2020-08-04
Remove support for clang/llvm 8.0

sunfishcode:
Remove old workarounds for clang/llvm 8.0, which we'll no longer support, for building WASI libc or WASI sdk releases.

Removing `--no-threads` was recently discussed in https://reviews.llvm.org/D76885 .

Also related is https://github.com/WebAssembly/wasi-libc/issues/212 in which we're planning to remove an LLVM 8.0 workaround from WASI libc.


sunfishcode:
Sorry about the PR race :-}.

PR #159, opened 2020-10-12
Update to latest wasi-libc.

sunfishcode:
This includes new-style command support.


pchickey (approved):

sbc100:
This features won't make it into the llvm-11-based SDK though right?  Users who want that will need to build the SDK with ToT ?

sunfishcode:
That's right. And actually, the changes needed to really use the new-style command support aren't in LLVM yet anyway.

sunfishcode:
The LLVM changes to support new-style commands are here: https://reviews.llvm.org/D89274.

PR #161, opened 2020-10-14
Update how wasi-sdk uses cmake

sunfishcode:
Change `CMAKE_SYSTEM_NAME` to `WASI`, use a `CMAKE_MODULE_PATH`,
remove `CACHE` usage, and fix the compiler autodetection code so that
the "compiler works" hacks are no longer needed, following the advice in
[this comment](https://gitlab.kitware.com/cmake/cmake/-/issues/19223#note_567327).

This is hopefully a step towards re-submitting WASI support to upstream cmake.


Makefile
```diff
@@ -55,6 +55,7 @@ clean:
 build/llvm.BUILT:
 	mkdir -p build/llvm
 	cd build/llvm && cmake -G Ninja \
+		-DCMAKE_MODULE_PATH=$(ROOT_DIR)/cmake \
```
sbc100, line 58: There is no cross building happening here so is this needed?

Makefile
```diff
@@ -138,9 +145,12 @@ LIBCXX_CMAKE_FLAGS = \
...
     --debug-trycompile
 
-    #-DCMAKE_STAGING_PREFIX= 
-
 build/libcxx.BUILT: build/llvm.BUILT build/compiler-rt.BUILT build/wasi-libc.BUILT
+	# Create an empty library so that cmake's configure checks which
+	# run the compiler to see if it produces working binaries succeed.
+	mkdir -p  $(PREFIX)/lib/clang/$(CLANG_VERSION)/lib/wasi
+	$(AR) cr $(PREFIX)/share/wasi-sysroot/lib/wasm32-wasi/libc++.a
+	$(AR) cr $(PREFIX)/share/wasi-sysroot/lib/wasm32-wasi/libc++abi.a
```
sbc100, line 153: This is a little gross..  but I don't know that right way to bootstrap here.

sbc100: Actually I think its slightly better to do `-DCMAKE_C_COMPILER_WORKS=ON` on the command line here.   Basically saying "we are bootstrapping here so skip compiler checks". 

Doing it in the wasi-sdk.cmake  globally was wrong but going it locally is not so bad IMHO.

wasi-sdk.cmake
```diff
@@ -16,23 +16,14 @@ endif()
 
 set(CMAKE_C_COMPILER ${WASI_SDK_PREFIX}/bin/clang${WASI_HOST_EXE_SUFFIX})
 set(CMAKE_CXX_COMPILER ${WASI_SDK_PREFIX}/bin/clang++${WASI_HOST_EXE_SUFFIX})
-set(CMAKE_AR ${WASI_SDK_PREFIX}/bin/llvm-ar${WASI_HOST_EXE_SUFFIX} CACHE STRING "wasi-sdk build")
-set(CMAKE_RANLIB ${WASI_SDK_PREFIX}/bin/llvm-ranlib${WASI_HOST_EXE_SUFFIX} CACHE STRING "wasi-sdk build")
-set(CMAKE_C_COMPILER_TARGET ${triple} CACHE STRING "wasi-sdk build")
-set(CMAKE_CXX_COMPILER_TARGET ${triple} CACHE STRING "wasi-sdk build")
-set(CMAKE_C_FLAGS "-v" CACHE STRING "wasi-sdk build")
-set(CMAKE_CXX_FLAGS "-v -std=c++11" CACHE STRING "wasi-sdk build")
```
sbc100, line 24: Weird.. why would we ever want `-v` to be the default cflags..

sunfishcode: Debugging code that got left behind...

Makefile
sunfishcode: Fixed.

Makefile
sunfishcode: I like that idea. I've now updated the PR to do this, and removed all the empty-archive hacks.

Makefile
```diff
@@ -155,6 +157,8 @@ build/libcxx.BUILT: build/llvm.BUILT build/compiler-rt.BUILT build/wasi-libc.BUI
 
 # Flags for libcxxabi.
 LIBCXXABI_CMAKE_FLAGS = \
+    -DCMAKE_C_COMPILER_WORKS=ON \
```
sbc100, line 160: Oh.. looks like these were already specified below!

sbc100 (approved):

PR #164, opened 2020-11-17
Build windows SDK with MSVC instead of gcc

### LINKED ISSUE #148: Windows builds require libcc_s_seh-1.dll to be available
The Windows Clang builds require `libcc_s_seh-1.dll`, which is present in the Git Bash environment (mingw based).  This is unexpected; I'm not sure why this dependency is coming in to clang itself.  We should fix this, as right now the generated wasi SDK for Windows only works if you have this DLL in your path from somewhere.

(I can take a look at some point, but if someone beats me to it please do)

vvuk:
By default, cmake was finding MSYS (or mingw, whichever) gcc in the path and building with it.  The resulting binaries therefore required `libgcc_s_seh-1.dll` to be available.

This executes a VS environment setup before the build, causing cmake to pick up on VC++ and generates a clean no-dependencies LLVM build (other than the Windows CRT).

Fixes #148

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,9 @@ jobs:
       - uses: actions/checkout@v1
         with:
           submodules: true
+      - name: Set up Visual Studio shell
+        uses: egor-tensin/vs-shell@v1
+        if: matrix.os == 'windows-latest'
       - name: Install ninja (Windows)
         run: choco install ninja
         if: matrix.os == 'windows-latest'
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           submodules: true
       - name: Set up Visual Studio shell
-        uses: egor-tensin/vs-shell@v1
+        uses: egor-tensin/vs-shell@6d88c01ad4b6b33190d261b45e241e462545652b
         if: matrix.os == 'windows-latest'
       - name: Install ninja (Windows)
         run: choco install ninja


vvuk:
cc @sbc100 @sunfishcode -- can't seem to request review on this

.github/workflows/main.yml
```diff
@@ -23,6 +23,9 @@ jobs:
       - uses: actions/checkout@v1
         with:
           submodules: true
+      - name: Set up Visual Studio shell
+        uses: egor-tensin/vs-shell@v1
```
sunfishcode, line 27: I'm not familiar with this space, but it's surprising to have a dependency on a third-party action that seems to be only 5 days old, to do something which seems like it should be fairly common. Is this the best way to do this?

If so, could you [add a SHA hash](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions)?

vvuk: Last time I looked at how to do this it was a lot of manual work to call out to the vcvars script, which is why I wasn't able to spend the time to figure it out.  I was very pleasantly surprised to discover this action that did the right setup, and did it in a way that made the tools visible in all subsequent shells. :)

vvuk: (I've also audited the action code -- it does Good Stuff (tm).  Added the sha hash of the v1 commit)

sunfishcode:
Sounds good, thanks!

PR #167, opened 2020-12-02
fix deb build

pchickey:
change makefile and deb_from_installation to invoke it with the new
build directory (changed in #143).

deb_from_installation will exit 0 if dpkg-deb missing so we dont try
to run it as part of `make package` on non-deb systems, but other errors
should cause the makefile to fail, which should catch regressions like
this sooner in the future.

### DIFF
--- a/deb_from_installation.sh
+++ b/deb_from_installation.sh
@@ -1,5 +1,16 @@
 #!/usr/bin/env sh
+set -x
+
+command -v dpkg-deb >/dev/null
+if [ $? -eq 0 ]
+then
+else
+    echo "required tool dpkg-deb missing. exiting"
+    exit 0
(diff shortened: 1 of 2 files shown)


sunfishcode (approved):
LGTM!

deb_from_installation.sh
```diff
@@ -1,5 +1,16 @@
 #!/usr/bin/env sh
+set -x
+
+command -v dpkg-deb >/dev/null
+if [ $? -eq 0 ]
+then
```
fgsch, line 6: if you change -eq to -ne you can remove the empy then.

pchickey: done, thanks!

PR #169, opened 2021-01-16
Add a testcase for getentropy

caspervonb:
This adds a basic testcase for `getentropy` which checks for errno and a sane result.

### DIFF
--- a/tests/general/getentropy.c
+++ b/tests/general/getentropy.c
@@ -1,0 +1,17 @@
+#include <assert.h>
+#include <unistd.h>
+
+int main() {
+  char buf[256] = {0};
+  int ret = getentropy(buf, 256);
+  assert(ret == 0);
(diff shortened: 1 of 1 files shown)


tests/general/getentropy.c
```diff
@@ -0,0 +1,17 @@
...
+#include <unistd.h>
+
+int main() {
+  char buf[256] = {0};
+  int ret = getentropy(buf, 256);
+  assert(ret == 0);
+
+  int sum = 0;
+  for (int i = 0; i < 256; i++) {
+    sum += buf[i];
```
jedisct1, line 11: Maybe use unsigned types and `^=` instead of `+=` to check for constant output (instead of specifically `0` output).

caspervonb:
Is the CI here broken?

PR #171, opened 2021-01-29
In clang 11.0 and newer, use a relative sysroot path.

### LINKED ISSUE #58: Can we have clang find the sysroot using a relative path?
Today we configure llvm with `-DDEFAULT_SYSROOT=/opt/wasi-sdk/share/wasi-sysroot` which means the resulting toolchain will work as expected only when installed in `/opt/wasi-sdk`.

Installing it anywhere else (e.g. just untaring it somewhere) will mean that one needs to pass the `--sysroot=..` flag to each compiler invocation. 

Perhaps there is some way we can find the default sysroot relative the compiler binary?

Followup from #57

sunfishcode:
This makes use of https://reviews.llvm.org/D76653, which is in clang 11
and newer, to fix #53. With this, clang can find the sysroot relative to
its own path.

Fixes #58.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,11 @@ build/llvm.BUILT:
 		-DLLVM_TARGETS_TO_BUILD=WebAssembly \
 		-DLLVM_DEFAULT_TARGET_TRIPLE=wasm32-wasi \
 		-DLLVM_ENABLE_PROJECTS="lld;clang;clang-tools-extra" \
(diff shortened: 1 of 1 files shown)


pchickey:
thank you @sbc100 for this upstream fix!

pchickey (approved):

sbc100:
Sweet!

PR #175, opened 2021-03-07
Update 

ghost:
Pulling into devices

### DIFF
--- a/tests/general/abort.c.stderr.expected
+++ b/tests/general/abort.c.stderr.expected
@@ -2,5 +2,5 @@ Error: failed to run main module `abort.c.---.wasm`
 
 Caused by:
     0: failed to invoke `_start`
-    1: wasm trap: unreachable, source location: @----
+    1: wasm trap: unreachable
        wasm backtrace:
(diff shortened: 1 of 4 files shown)


tests/general/abort.c.stderr.expected
```diff
@@ -2,5 +2,5 @@ Error: failed to run main module `abort.c.---.wasm`
 
 Caused by:
     0: failed to invoke `_start`
-    1: wasm trap: unreachable, source location: @----
```
sunfishcode, line 5: Can you also update abort.c.stderr.expected.filter to filter out this part?

tests/general/ctors_dtors.c.stdout.expected
```diff
@@ -5,6 +5,6 @@ hello main
 goodbye main
 hello another_from_atexit
 hello from_atexit
-hello from_destructor65535
 hello from_destructor
+hello from_destructor65535
```
sunfishcode, line 9: In native code, `from_destructor65535` runs before `from_destructor`, and I believe the existing expected output is correct. And with newer LLVM releases that have [this commit](https://github.com/llvm/llvm-project/commit/d865437d9c82f814dd3b0d0e60042e2c176e201f) in them, that's what wasi-sdk does too.

Assuming that this is the underlying cause of the difference here, would it be practical for you to user a newer LLVM here?

sunfishcode:
This PR's author is now listed as @ghost.

sunfishcode:
This PR's author is now listed as @ghost.

PR #177, opened 2021-03-22
Add tests for `clock`, `times`, and `getrusage`.

sunfishcode:
This is a basic test for https://github.com/WebAssembly/wasi-libc/pull/241.

### DIFF
--- a/tests/general/clocks.c
+++ b/tests/general/clocks.c
@@ -1,0 +1,43 @@
+#include <time.h>
+#include <sys/resource.h>
+#include <sys/times.h>
+#include <assert.h>
+
+static void test_clock(void) {
+    clock_t a = clock();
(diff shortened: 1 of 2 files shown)


ricochet:
Should this update wasi-libc? That is usually in its own PR.

sunfishcode:
> Should this update wasi-libc? That is usually in its own PR.

The update to wasi-libc is needed to pick up b9b64a695a35273f92574acc6270f0e059be3dca, which is needed to make these tests pass.

sunfishcode:
Ah, it turns out that wasi-libc has already updated, so I've now dropped that change.

sunfishcode:
This is just adding a few simple tests, so I think we can merge this.

PR #178, opened 2021-04-16
use llvm 12.0.0 release

pchickey:


sbc100 (approved):
Exciting!

jedisct1:
> Exciting!

Not so much; it was tagged even though 13 known regressions since LLVM 11 are still open.

https://bugs.llvm.org/show_bug.cgi?id=48902
https://twitter.com/andy_kelley/status/1382443048571785216

I wouldn't rush.

pchickey:
Of those regressions, it only looks like 2 of the regressions are miscompilations in InstCombine. I don't know for sure if those could affect wasm32-wasi. I'm inclined to go ahead and release on top of 12.0.0 unless @sbc100 can indicate that a 12.0.1 is coming shortly.

sbc100:
In emscripten we have seen some regressions due to the new pass manager.. but I think we planning on simply accepting them.

sbc100:
I'm not following the plans for a 12.0.1 release.

pchickey:
Thanks. Sounds like we should just go ahead with a release based on 12.0.0 and do another release if or when patches come out.

Issue #181, opened 2021-05-26
`set(CMAKE_SYSTEM_NAME WASI)` issues on macOS host


axic:
It is great to have [`wasi.cmake`](https://github.com/WebAssembly/wasi-sdk/blob/main/wasi-sdk.cmake) and we use it in several projects ([ethash](https://github.com/chfast/ethash),[evmone](https://github.com/ethereum/evmone)), however during integration we have noticed one shortcoming stemming from the `set(CMAKE_SYSTEM_NAME WASI)` setting.

CMake will try to find [`Platform/WASI.cmake`](https://github.com/WebAssembly/wasi-sdk/blob/main/cmake/Platform/WASI.cmake), but if missing, it will fall back to the host system. This works okay on Linux systems, but fails on macOS as that will add compiler settings only available/valid for macOS targets.

There are two possible fixes I have found:
1. Copy the platform file to CMake
2. Change/override the `CMAKE_SYSTEM_NAME` to `Linux`

The downside of option 1 is that it requires change to the installed cmake (copying a new file), while using the toolchain file works without any host changes.

We have decided to override the system name in our files, but I wonder if it would make sense to just change it to Linux in this repository. Any opinions? I can submit a PR if that is there's agreement for this change.

If there's any other options on top of the above, please let me know.

sbc100:
I think that this can be made to work without installing into the system location.

We already include both  `wasi.cmake` and `cmake/Platform/WASI.cmake` in this repo, so I think it just a matter of setting up the cmake paths correctly to have it pick up the `cmake/Platform/WASI.cmake` file correctly.

sbc100:
(I'm not sure what the exact answer is.. but the existence of `cmake/Platform/WASI.cmake` is good hint)

axic:
Hmm, for some reason on our end the platform file was not picked up. Do you have a hint how to set an alternative platform path?

sbc100:
Looks like its `CMAKE_MODULE_PATH`. 

Looking at the emscripten cmake plugin it looks like the toolchain file injects:
https://github.com/emscripten-core/emscripten/blob/3c0587f94c2b9b0e29ea6a4054a1b8ea9a148f9d/cmake/Modules/Platform/Emscripten.cmake#L77

I wonder if that is pattern we can/should copy?

axic:
[...]

PR #182, opened 2021-05-26
wasi-sdk.cmake: Add wasi-sdk/cmake to CMAKE_MODULE_PATH

sbc100:
Fixes: #181

### DIFF
--- a/wasi-sdk.cmake
+++ b/wasi-sdk.cmake
@@ -3,6 +3,10 @@
 # This is arbitrary, AFAIK, for now.
 cmake_minimum_required(VERSION 3.4.0)
 
+# Until Platform/WASI.cmake is upstream we need to inject the path to it
+# into CMAKE_MODULE_PATH.
+list(APPEND CMAKE_MODULE_PATH "${WASI_SDK_PREFIX}/cmake")
+
 set(CMAKE_SYSTEM_NAME WASI)
 set(CMAKE_SYSTEM_VERSION 1)
 set(CMAKE_SYSTEM_PROCESSOR wasm32)


wasi-sdk.cmake
```diff
@@ -3,6 +3,10 @@
 # This is arbitrary, AFAIK, for now.
 cmake_minimum_required(VERSION 3.4.0)
 
+# Until Platform/WASI.cmake is upstream we need to inject the path to it
+# into CMAKE_MODULE_PATH.
+list(APPEND CMAKE_MODULE_PATH "${WASI_SDK_PREFIX}/cmake")
```
axic, line 8: I am using the wasi-sdk release tarball, which has `share/cmake/wasi-sdk.cmake` and does not contain the platform. Should the platform file be included at `share/cmake/Platform/WASI.cmake`, the following change would be needed:

```suggestion
list(APPEND CMAKE_MODULE_PATH "${WASI_SDK_PREFIX}/share/cmake")
```

Doing these manually locally does seem to make it work.

sunfishcode: @sbc100 Does this change make sense?

sbc100: We should probably add some kind of minimal cmake test to the CI here too.

axic: Just to confirm, the release tarball (for 12.0 at least) does not contain the platform file. I placed it to `SDK/share/cmake/Platform/WASI.cmake` to mirror directory layout here.

sunfishcode: As @sbc100 mentioned, is there a way we can add a minimal test here to the CI to ensure that this keeps working?

arvid-norlander:
Any updates on this?

abrown:
> Any updates on this?

I think this PR has long been forgotten, but from glancing at it it seems like a reasonable change. Are you interested in finishing this out? I think what is left are the "add this file to the release packages" and "check that this works in CI" discussed above. I'll be glad to merge this.

arvid-norlander:
I certainly have an interest in seeing this fixed, so that I don't have to carry a local patch. To spare me having to hunt around: how/where are the release packages built?

arvid-norlander:
Quoting @sbc100  above:
>     2. Is this file used in-place? (i.e. do we use this from the git checkout before install?)  If so we have two different ways we might want to find WASI.cmake and we need to fix it so it work both in-tree and out-of-tree.

I don't know the answer to this. It is unfortunate that the files are in different location in the release and the source. Perhaps one could be moved to the other? For example using `share/cmake` in the source as well?

On the other hand, it seems that the Makefile explicitly sets `CMAKE_MODULE_PATH` in several places, so perhaps this is not an issue?

arvid-norlander:
> > Any updates on this?
> 
> I think this PR has long been forgotten, but from glancing at it it seems like a reasonable change. Are you interested in finishing this out? I think what is left are the "add this file to the release packages" and "check that this works in CI" discussed above. I'll be glad to merge this.

I have made a [branch in a fork of this repo](https://github.com/arvid-norlander/wasi-sdk/tree/cmake_module_path) for this. I assume a separate PR will be needed for it.

However, I'm not sure how you would check that this works in CI. Some advise would be appreciated on what a suitable CI check for this would be. Maybe we could need to test building with release SDK, and check that WASI is set in cmake? However, looking at the current test suite, it does not seem to be setup for using cmake currently, and this might be a larger effort.

abrown:
Hm, yeah, I see what you mean about no current CMake support in the CI infrastructure. But maybe the place to inject that is here:

https://github.com/WebAssembly/wasi-sdk/blob/00cb3b2e5ffc9c747311f5d22cc9cf49f3be12e5/tests/run.sh#L32-L40

I don't use CMake very often but is there a way to run those compile-only tests with CMake? I'm wondering if there is a single-line CMake command that can be slotted in for `$CC` or `$CXX`, but maybe that requires the entire CMake project infrastructure?

arvid-norlander:
> I don't use CMake very often but is there a way to run those compile-only tests with CMake? I'm wondering if there is a single-line CMake command that can be slotted in for `$CC` or `$CXX`, but maybe that requires the entire CMake project infrastructure?

Not sure what you mean, but we could set up a small cmake project that verifies that WASI is set (which will be a good indicator that the platform file exists). Additionally we should probably build a small test program, and preferably execute it too. Assuming cmake is already installed in your CI images, I could definitely cook up something that does this (I use cmake daily). The only unknown is how to execute a built program, as my use case at work (where we use cmake) is using an embedded wasm engine (WAMR).

arvid-norlander:
What version of cmake do you have in your CI? So I can figure out what minimum version to put in the cmake file. I doubt I will need anything super-modern, so targeting what is in the current Ubuntu LTS is probably a good start?

arvid-norlander:
I have implemented this, and I will open a separate PR

abrown:
This is superseded by #335.

PR #188, opened 2021-08-09
Update to latest LLVM and wasi-libc

sunfishcode:
Among other things, this pullls in https://github.com/llvm/llvm-project/commit/1d445a6e7679cc188fd051f7397b7d9ca8ce4f10, which is needed by https://github.com/bytecodealliance/wizer/pull/8 .


sbc100:
Don't we want to pin to specific llvm releases here?  (or is this the llvm 13 branch point?)

sunfishcode:
Ah, no, this isn't the LLVM 13 branch point. This is just pulling in a recent revision so that we get the above-mentioned patch to fix  bytecodealliance/wizer#8.

wask-sdk has previously used LLVM release revisions, but I don't know that we need to. If our own tests pass, and LLVM's buildbots are reasonably healthy, and we have a real-world use case that would benefit from a recent fix, it seems advantageous to pull in the new revision.

sbc100:
I think it would be nice to stick to released versions but I wont block this.   IIUC the workaround for bytecodealliance/wizer#8 is to just run `wasm-opt` which does the de-duplication so I'm surprised this is urgent for that use case.  

lgtm though if you feel like it worth it.

sbc100 (approved):

sunfishcode:
This is superceded by #197.

PR #189, opened 2021-08-11
Add missing dependency between compiler-rt and wasi-libc

glandium:


sbc100:
Can you explain why this is needed?  Does compiler-rt depend on libc?

glandium:
> Does compiler-rt depend on libc?

It does. (it includes std headers)

sbc100 (approved):

PR #197, opened 2021-10-07
Update to LLVM 13.0

sunfishcode:
This also changes the CI to build on bionic instead of xenial, which has a newer version of Python.

This pullls in llvm/llvm-project@1d445a6, which is needed by bytecodealliance/wizer#8 .

This supercedes #188, and updates to a released version.


lygstate:
Stable 13.0 are released

PR #207, opened 2021-11-19
Setup ninja verbose option at the workflow scripts

lygstate:
When building wasi-sdk locally, we doesn't need so verbose output that not easily distinguish
the error command from the normal command. So refactoring -v option into workflow scripts

Signed-off-by: Yonggang Luo <luoyonggang@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ build/llvm.BUILT:
 		     -DDEFAULT_SYSROOT=$(PREFIX)/share/wasi-sysroot) \
 		-DLLVM_INSTALL_BINUTILS_SYMLINKS=TRUE \
 		$(LLVM_PROJ_DIR)/llvm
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -4,6 +4,9 @@
 ROOT_DIR=${CURDIR}
 LLVM_PROJ_DIR?=$(ROOT_DIR)/src/llvm-project
 
+# Setup ninja verbose option at the very beginning for easily comment it out
+NINJA_FLAGS:=$(NINJA_FLAGS) -v
```
sbc100, line 8: How about we set `NINJA_FLAGS=-v` in the CI scripts and then we can avoid having this as the default at all?

lygstate: OK with that

sbc100 (approved):

PR #103, opened 2020-02-21
WIP: Re-enable windows build in CI

sbc100:

PR #208, opened 2021-11-20
Split github actions building for windows(x86/x64) separately

lygstate:
Also the tests now can running on Windows and OSX fine

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,17 +18,10 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
-          - windows-latest
     steps:
       - uses: actions/checkout@v1
         with:
           submodules: true
-      - name: Set up Visual Studio shell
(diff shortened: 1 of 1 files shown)


lygstate:
This may also close https://github.com/WebAssembly/wasi-sdk/pull/103

sbc100:
Is there much value in supporting both 32 and 64-bit windows?  Could we just build and ship one of them because either (a) there are not enough 32-bit windows users out there, or (b) the 64-bit windows users can just run the 32-bit binaries?   I guess that latter might not work since IIRC LTO link steps can exceed 4Gb memory limits...

lygstate:
> Is there much value in supporting both 32 and 64-bit windows? Could we just build and ship one of them because either (a) there are not enough 32-bit windows users out there, or (b) the 64-bit windows users can just run the 32-bit binaries? I guess that latter might not work since IIRC LTO link steps can exceed 4Gb memory limits...

As LLVM still release x86 version, and on Windows, x86 are still broadly used

PR #234, opened 2022-05-10
Build libcxx and libcxxabi via llvm-project/runtimes

glandium:
This is the new method to build runtimes as of LLVM 12, and the previous
standalone method is deprecated since LLVM 14 and unsupported since
LLVM 15 (https://github.com/llvm/llvm-project/commit/6f17768e11480063f4c2bcbeea559505fee3ea19)

Building libcxx and libcxxabi separately doesn't currently work in LLVM
because of some wrong dependencies in tests CMakeLists.txt in that
configuration. They could both be built in one go with
-DLLVM_ENABLE_RUNTIMES="libcxx;libcxxabi", but that wouldn't be
compatible with older versions of LLVM. Tests are however not needed, so
we just disable them.

LIBCXXABI_LIBCXX_INCLUDES being only used when building standalone, we
need to add the path manually to C++ flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ build/compiler-rt.BUILT: build/llvm.BUILT build/wasi-libc.BUILT
 	cp -R $(ROOT_DIR)/build/llvm/lib/clang $(BUILD_PREFIX)/lib/
 	touch build/compiler-rt.BUILT
 
-# Flags for libcxx.
+# Flags for libcxx and libcxxabi.
 LIBCXX_CMAKE_FLAGS = \
     -DCMAKE_C_COMPILER_WORKS=ON \
     -DCMAKE_CXX_COMPILER_WORKS=ON \
@@ -136,11 +136,13 @@ LIBCXX_CMAKE_FLAGS = \
     -DCMAKE_STAGING_PREFIX=$(PREFIX)/share/wasi-sysroot \
     -DLLVM_CONFIG_PATH=$(ROOT_DIR)/build/llvm/bin/llvm-config \
     -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
+    -DCXX_SUPPORTS_CXX11=ON \
     -DLIBCXX_ENABLE_THREADS:BOOL=OFF \
     -DLIBCXX_HAS_PTHREAD_API:BOOL=OFF \
     -DLIBCXX_HAS_EXTERNAL_THREAD_API:BOOL=OFF \
     -DLIBCXX_BUILD_EXTERNAL_THREAD_LIBRARY:BOOL=OFF \
     -DLIBCXX_HAS_WIN32_THREAD_API:BOOL=OFF \
+    -DLLVM_COMPILER_CHECKED=ON \
     -DCMAKE_BUILD_TYPE=RelWithDebugInfo \
     -DLIBCXX_ENABLE_SHARED:BOOL=OFF \
     -DLIBCXX_ENABLE_EXPERIMENTAL_LIBRARY:BOOL=OFF \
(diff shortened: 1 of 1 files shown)


glandium:
And of course, windows builds fail because of path garbling from a msys->win32 transition...

Makefile
```diff
@@ -146,11 +146,13 @@ LIBCXX_CMAKE_FLAGS = \
     -DLIBCXX_ENABLE_EXPERIMENTAL_LIBRARY:BOOL=OFF \
     -DLIBCXX_ENABLE_EXCEPTIONS:BOOL=OFF \
     -DLIBCXX_ENABLE_FILESYSTEM:BOOL=OFF \
+    -DLIBCXX_INCLUDE_TESTS=OFF \
     -DLIBCXX_CXX_ABI=libcxxabi \
     -DLIBCXX_CXX_ABI_INCLUDE_PATHS=$(LLVM_PROJ_DIR)/libcxxabi/include \
     -DLIBCXX_HAS_MUSL_LIBC:BOOL=ON \
     -DLIBCXX_ABI_VERSION=2 \
     -DWASI_SDK_PREFIX=$(BUILD_PREFIX) \
+    -DUNIX:BOOL=ON \
```
sbc100, line 155: This looks a little odd.. maybe not part of final PR?

glandium: No, it's there on purpose. The same is already there in the libcxxabi build, somehow wasn't required for libcxx but is when using runtimes/

Makefile
```diff
@@ -201,9 +205,11 @@ build/libcxxabi.BUILT: build/libcxx.BUILT build/llvm.BUILT
 	cd build/libcxxabi && cmake -G Ninja $(LIBCXXABI_CMAKE_FLAGS) \
 		-DCMAKE_SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
 		-DCMAKE_C_FLAGS="$(DEBUG_PREFIX_MAP)" \
-		-DCMAKE_CXX_FLAGS="$(DEBUG_PREFIX_MAP)" \
+		-DCMAKE_CXX_FLAGS="$(DEBUG_PREFIX_MAP) -I$(BUILD_PREFIX)/share/wasi-sysroot/include/c++/v1" \
```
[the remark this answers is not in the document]
glandium: It would, if we're okay dumping support for LLVM < 12.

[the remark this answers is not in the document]
sbc100: I think supporting the current version of llvm plus 2 prior releases should be enough.  Supporting llvm 11 would mean supporting 3 prior releases.

WDYT @sunfishcode ?  @pchickey ?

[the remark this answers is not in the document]
sunfishcode: I think dropping LLVM < 12 in the wasi-sdk build scripts sounds fine at this point. People using older LLVMs can either switch to using prebuilt sysroots, or just stick to wasi-sdk-15 for a while, which we just released.

Makefile
```diff
@@ -126,21 +126,24 @@ build/compiler-rt.BUILT: build/llvm.BUILT build/wasi-libc.BUILT
...
     -DCMAKE_C_COMPILER_WORKS=ON \
     -DCMAKE_CXX_COMPILER_WORKS=ON \
     -DCMAKE_AR=$(BUILD_PREFIX)/bin/ar \
     -DCMAKE_MODULE_PATH=$(ROOT_DIR)/cmake \
+    -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
     -DCMAKE_TOOLCHAIN_FILE=$(ROOT_DIR)/wasi-sdk.cmake \
     -DCMAKE_STAGING_PREFIX=$(PREFIX)/share/wasi-sysroot \
     -DLLVM_CONFIG_PATH=$(ROOT_DIR)/build/llvm/bin/llvm-config \
     -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
+    -DCXX_SUPPORTS_CXX11=ON \
```
sbc100, line 140: Why is this needed?

Makefile
```diff
@@ -126,21 +126,24 @@ build/compiler-rt.BUILT: build/llvm.BUILT build/wasi-libc.BUILT
...
 	touch build/compiler-rt.BUILT
 
-# Flags for libcxx.
+# Flags for libcxx and libcxxabi.
 LIBCXX_CMAKE_FLAGS = \
     -DCMAKE_C_COMPILER_WORKS=ON \
     -DCMAKE_CXX_COMPILER_WORKS=ON \
     -DCMAKE_AR=$(BUILD_PREFIX)/bin/ar \
     -DCMAKE_MODULE_PATH=$(ROOT_DIR)/cmake \
+    -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
```
sbc100, line 135: I don't think we need to add this here.  You can do `NINJA_FLAGS=-v` if you want this I think?

Makefile
sbc100: Oh I see this is a result of the merging..

Makefile
sbc100: Hmm.. actually this is already set down on line 138.. I think we should remove it but I guess that is a separate issue.

sbc100: In any case this line is a duplicate of line 138 I think and so can be removed?

sbc100 (commented):
Looks reasonable to me.  How much testing of libc++/libc++abi do we have in CI here?

Makefile
glandium: Err, I missed that there was one already.

sunfishcode:
We don't have a lot of C++ testing in CI, however my sense is that this PR looks safe enough that we can merge it and be ok here.

sunfishcode:
Thanks, @glandium!

PR #248, opened 2022-08-04
Add llvm-mc

TerrorJack:
This PR adds `llvm-mc` to `wasi-sdk`. We intend to use `llvm-mc` instead of `clang` as a `wasm32` assembler, and this would allow us to pass `-no-type-check` to bypass the rather error-prone `wasm32` assembly type checker.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ build/llvm.BUILT:
 		install-clang-tidy \
 		install-clang-apply-replacements \
 		install-lld \
+		install-llvm-mc \
 		install-llvm-ranlib \
 		install-llvm-strip \
 		install-llvm-dwarfdump \


sbc100:
If I land https://reviews.llvm.org/D131217 then hopefully this won't be needed?

TerrorJack:
I believe this PR will still be needed, given that change needs to propagate to an LLVM release and then to here, which takes a very long time.

TerrorJack:
ping?

sbc100:
I was going to suggest we add a comment mentioning why this is needed and as a reminder to remove it once the MC bug has been fixed and/or the clang passthrough added.. but maybe its useful to have llvc-mc directly accessible anyway, what do you think @sunfishcode .. do we want this even in the long term?

sunfishcode:
I think we can ship llvm-mc, as it's useful for some users.

sunfishcode:
I guess the main argument against shipping it is that it's largely redundant. And that's true, though llvm-mc does have a different command-line with a lot of its own flags, and I don't think it makes sense to add all of the flags to clang.

sbc100:
> I think we can ship llvm-mc, as it's useful for some users.

I'm curious what kind of users?  Can you give an example?  (In this case its just a workaround bugs in MC/clang).   I'm not against landing this just curious.

sbc100:
OK.. I guess that "we don't want to add all the MC flags to clang" argument make sense to me.  I didn't know that was a non-goal of clang.

sbc100 (approved):

sunfishcode:
I don't know that it's an explicit non-goal, but if supporting an llvm-mc flag in `clang -Wa` requires a PR like https://reviews.llvm.org/D131217 where the driver has to know about the flag, just to pass it on, that's a lot of clutter if we did it for all off llvm-mc's flags, so it seems better to just ship llvm-mc for people who need all those flags.

PR #261, opened 2022-12-02
Update to the latest wasi-libc.

sunfishcode:
This adds the `__errno_location` function and the change to `readdir` to call `stat` if `d_ino` is 0.


sunfishcode:
It also contains the change to handle zero inode values in `fd_readdir`.

sunfishcode:
This is just updating to the latest wasi-libc.

PR #265, opened 2022-12-03
Use msys2 clang toolchain to build windows distribution

### LINKED ISSUE #264: Build windows version with msys2 clang64 toolchain instead of msvc?
It seems the msvc builds has been failing on recent PRs. Recently I had success in building wasi-sdk using msys2 clang64 toolchain on Windows, and the resulting binaries do not contain any dll dependencies in the msys2 environment so they can be used in regular Windows environment too. If this sounds like something useful to your folks, I can open a PR to change the build logic for windows :)

TerrorJack:
Closes #264.

Note: still a WIP, will undraft and ping for review after:

- [x] Both x64/x86 builds are ready
- [x] There's basic testing that msys2 dll dependencies don't leak. `make check` won't serve this purpose since it needs to be run in a msys2 shell, for now I believe `clang --version` and `wasm-ld --version` should suffice since they are the largest executables anyway.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,42 +42,45 @@ jobs:
 
   winbuild:
     name: Windows Build
-    runs-on: windows-2019
+    runs-on: windows-latest
     strategy:
       fail-fast: false
       matrix:
-        arch:
-          - x64
-          - x86
-    env:
-      MSYS2_PATH_TYPE: inherit
-      MSYSTEM: MINGW64
-      CC: cl
-      CXX: cl
-      NINJA_FLAGS: -v
+        include:
+          - arch: x64
+            sys: clang64
+            env: clang-x86_64
+          - arch: x86
+            sys: clang32
+            env: clang-i686
     steps:
-      - uses: actions/checkout@v1
+      - uses: msys2/setup-msys2@v2
         with:
+          install: >-
+            base-devel
+            git
+            mingw-w64-${{ matrix.env }}-cmake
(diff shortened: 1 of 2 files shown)


TerrorJack:
@sunfishcode Ready for review.

sunfishcode:
Does the switch from windows-2019 to windows-latest (which is windows-2022) mean that we lose any support or older Windows versions in our distributed artifacts?

TerrorJack:
> Does the switch from windows-2019 to windows-latest (which is windows-2022) mean that we lose any support or older Windows versions in our distributed artifacts?

The binaries will run fine on Windows 10 and later versions that ships ucrt in system32. If you wish I can also switch the build platform to windows-2019, though it doesn't provide any meaningful assurance in compatability, since the build process does not use systemwide visual studio installation at all.

TerrorJack:
Untitled

I just started a Windows 7 SP1 x64 virtual machine, and the new artifacts work there as long as [ucrt](https://support.microsoft.com/en-us/topic/update-for-universal-c-runtime-in-windows-c0514201-7fe6-95a3-b0a5-287930f3560c) is installed. I believe that's a reasonable lowest bound on supported Windows versions. And this is not a regression from the old msvc-produced artifacts, those also require ucrt to be manually installed (and additionally msvcp140).

sunfishcode:
Great, thanks for investigating that! I think we can switch to this for our next release, which I'm hoping to put out soon.

PR #271, opened 2022-12-19
Add docker image that can be used to build projects with wasi-sdk sysroot

### LINKED ISSUE #176: Publish an official Docker image
It would be nice to have an official wasi-sdk Docker image for use in CI; the [current images I see on Docker Hub are unofficial](https://hub.docker.com/search?q=wasi-sdk&type=image).

anuraaga:
Fixes #176 

Based on the issue it seems there is interest in this with some help, so sent a PR to add building and publishing of a wasi-sdk docker image. emscripten provides an official emsdk docker image to make it easier to build existing projects to webassembly, for example [wabt](https://github.com/WebAssembly/wabt/blob/main/.github/workflows/build.yml#L68) uses it to build their emscripten artifact. A similar docker image for wasi-sdk can allow C++ projects to more easily provide webassembly for use outside the browser - for example, building wabt and binaryen to wasi will allow full webassembly toolchains without the need for installing multiple tools or C toolchains.

The docker image is built from main pushes with `main` and the github sha as tags, and for tag pushes with the tag name. You can see what it looks like in my fork https://github.com/anuraaga/wasi-sdk/pkgs/container/wasi-sdk

Using ghcr allows publishing docker with no worries of account or secret management. In the future, it could also be published to Docker Hub if desired but I personally don't see much point in OSS projects doing so anymore.

I have confirmed that wabt (cmake project) and re2 (make project) build fine with this image.

This approach does not use the LLVM built with wasi-sdk but instead installs packages and configures environment variables to use the wasi sysroot / target. This is to support multiplatform docker images - with arm64 becoming more popular it is important to provide images that work with it. While it's possible to build arm64 with qemu on GitHub Actions, for a monster like LLVM it's impractical.

autotools, cmake and ninja are also included to represent a common set of build tools used by C++ projects out there - the more included the bigger the image gets, but this feels like good coverage. 

Let me know what you think


.gitignore
```diff
@@ -1,2 +1,4 @@
 build
 dist
+.idea
+.vscode
```
sbc100, line 4: I'd rather not add this kind of thing here.  If you use vscode you are probably better off adding this to your global `.gitignore`, no?

sdk.Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+    # Temporarily install to setup apt repositories
+    apt-get install -y curl gnupg && \
+\
+    curl -sS https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor > /etc/apt/trusted.gpg.d/llvm.gpg && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/llvm.gpg] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${LLVM_VERSION} main" >> /etc/apt/sources.list.d/llvm.list && \
+    echo "deb-src [signed-by=/etc/apt/trusted.gpg.d/llvm.gpg] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${LLVM_VERSION} main" >> /etc/apt/sources.list.d/llvm.list && \
+\
+    curl -sS https://apt.kitware.com/keys/kitware-archive-latest.asc | gpg --dearmor > /etc/apt/trusted.gpg.d/kitware.gpg && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/kitware.gpg] https://apt.kitware.com/ubuntu/ jammy main" >> /etc/apt/sources.list.d/kitware.list && \
+    echo "deb-src [signed-by=/etc/apt/trusted.gpg.d/kitware.gpg] https://apt.kitware.com/ubuntu/ jammy main" >> /etc/apt/sources.list.d/kitware.list && \
```
sbc100, line 30: Surly the stock unbuntu version of cmake is good enough?

sdk.Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+ENV LLVM_VERSION 15
+
+# Install build toolchain including clang, ld, make, autotools, ninja, and cmake
+RUN apt-get update && \
+    # Temporarily install to setup apt repositories
+    apt-get install -y curl gnupg && \
+\
+    curl -sS https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor > /etc/apt/trusted.gpg.d/llvm.gpg && \
+    echo "deb [signed-by=/etc/apt/trusted.gpg.d/llvm.gpg] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${LLVM_VERSION} main" >> /etc/apt/sources.list.d/llvm.list && \
+    echo "deb-src [signed-by=/etc/apt/trusted.gpg.d/llvm.gpg] http://apt.llvm.org/jammy/ llvm-toolchain-jammy-${LLVM_VERSION} main" >> /etc/apt/sources.list.d/llvm.list && \
```
sbc100, line 26: Can we maybe just use the llvm that ships with jammy?  Is 13.0 recent enough?

sdk.docker.cmake
```diff
@@ -0,0 +1,27 @@
+# Cmake toolchain description file for the wasi-sdk docker image
```
sbc100, line 1: Can you move these three files into a `docker` subdirectory perhaps?

sdk.docker.cmake
```diff
@@ -0,0 +1,27 @@
...
+
+set(CMAKE_SYSTEM_NAME WASI)
+set(CMAKE_SYSTEM_VERSION 1)
+set(CMAKE_SYSTEM_PROCESSOR wasm32)
+set(triple wasm32-wasi)
+
+set(CMAKE_C_COMPILER /usr/bin/clang-15)
+set(CMAKE_CXX_COMPILER /usr/bin/clang++-15)
+set(CMAKE_AR /usr/bin/llvm-ar-15)
+set(CMAKE_RANLIB /usr/bin/llvm-ranlib-15)
```
sbc100, line 17: Given that you set CC/CXX etc in the environment maybe this is redundant here?

sdk.Dockerfile
anuraaga: Sure, was initially using debian where it was too old to support the `CMAKE_TOOLCHAIN_FILE` environment variable, but ubuntu's stock is fine

sdk.Dockerfile
anuraaga: It would worry me to use a version of LLVM older than what's published with wasi-sdk for potential incompatibilities, currently it looks like wasi-sdk is on 15.0.6

sdk.docker.cmake
anuraaga: With cmake, CC/CXX do get sourced from env variables, but AR RANLIB still don't unfortunately. It makes the state clearest I think to have it all in one place here rather than relying on environment for some but not all

anuraaga (commented):
Thanks @sbc100 - applied some cleanups and added some responses

docker/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+FROM ubuntu:22.04 as dist
+
+ADD dist/wasi-sdk-*.*-linux.tar.gz /
+
+# Move versioned folder to unversioned to using bash glob to allow
+# this file to be independent of major version number.
+RUN mv /wasi-sdk-* /wasi-sdk
+
+# Use ubuntu to use official repository with newer cmake packages
+FROM ubuntu:22.04
```
sbc100, line 15: can you explain the two different "FROM ubuntu" lines above?

docker/Dockerfile.dockerignore
```diff
@@ -0,0 +1 @@
+# Overrides top level .dockerignore to allow adding files to the image
```
sbc100, line 1: Can you explain this?  What top level file is being refereed too?    Can this file simply be call `.dockerignore` or does it need the longer name?

docker/sdk.cmake
```diff
@@ -0,0 +1,27 @@
+# Cmake toolchain description file for the wasi-sdk docker image
```
sbc100, line 1: Why not call this `wasi-sdk.cmake`, just like the file in the directory above?

.github/workflows/main.yml
```diff
@@ -7,6 +7,7 @@ on:
     branches:
       - main
   pull_request:
+  workflow_dispatch:
```
sbc100, line 10: Why this change?

anuraaga: It allows manually triggering the workflow from UI, notably the workflow can be run from a branch on a fork to test changes to the workflow before sending a PR. I can revert this if needed, but I think it makes development of this file safer without any real downside

docker/Dockerfile
anuraaga: Tried extending the comment above the first one, it's a multi-stage dockerfile, we don't need the entire SDK so we extract in a prepatory stage to reduce docker image size

docker/Dockerfile.dockerignore
anuraaga: It's the `.dockerignore` file in the root of the repository, which excludes all files since they aren't used in that build. This build does need files added.

The naming scheme is required to be like this for docker to pick it up as an override

anuraaga (commented):
Thanks @sbc100 applied some fixes and clarifications

docker/Dockerfile.dockerignore
sbc100: Can we instead just remove the top level one.. I don't know why that exists.

docker/Dockerfile
sbc100: nice.

.github/workflows/main.yml
sbc100: great ... seems like that should be the default.

docker/Dockerfile.dockerignore
anuraaga: Thanks @sbc100 - I've removed both. I suspect one motivation may have been the size of the submodules when initializing docker, but even on a slowish CI machine it seems to be just 4s so seems like an overoptimization

https://github.com/anuraaga/wasi-sdk/actions/runs/3728956980/jobs/6324422776#step:9:174

sbc100: Can you explain more?   Surly docker should not need to read the sub modules directory should it?  Doesn't it just include what we tell it to?

sbc100 (approved):
lgtm % comment/question.

Not sure if @sunfishcode or @pchickey have more to add?

docker/Dockerfile.dockerignore
anuraaga: Unfortunately no, docker begins by sending the entire context to the daemon - usually this is the current directory, `.` and is recursive. So to prevent transferring things during this step you need to use dockerignore, the values in ADD statements in the Dockerfile don't influence it.

sunfishcode:
I don't have anything to add here; if people would find this useful, let's do it.

docker/Dockerfile.dockerignore
sbc100: > Unfortunately no, docker begins by sending the entire context to the daemon - usually this is the current directory, `.` and is recursive. So to prevent transferring things during this step you need to use dockerignore, the values in ADD statements in the Dockerfile don't influence it.

That seems rather unfortunate!  So even if I only add one tiny file to the image the docker create command will be `O(size of everything my my CWD)`.   I did not know that.  

In that case maybe lets leave the top level `.dockerignore` but have it only ignore the large directories such as `dist`.

sbc100: Alternatively, can we run docker from the subdirectory and `ADD ../dist`?

anuraaga: Nope, docker won't allow adding files that are't in the context.

> In that case maybe lets leave the top level .dockerignore but have it only ignore the large directories such as dist.

Since this image needs dist (to be able to package it), I think it would result in the two dockerignore files again, one ignoring dist at top level and the other not. Just to confirm that's ok?

sbc100: No, I think we should just have one top level dockerignore, and it should just specify the stuff we don't want (i.e. the submodules).

anuraaga: Thanks @sbc100 - added a top level .dockerignore that excludes /src, the submodules folder.

docker/Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+ADD cmake/Platform/WASI.cmake /usr/share/cmake/Modules/Platform/WASI.cmake
+
+ENV CC clang-${LLVM_VERSION}
+ENV CXX clang++-${LLVM_VERSION}
+ENV LD wasm-ld-${LLVM_VERSION}
+ENV AR llvm-ar-${LLVM_VERSION}
+ENV RANLIB llvm-ranlib-${LLVM_VERSION}
+
+ENV CFLAGS --target=wasm32-wasi --sysroot=/wasi-sysroot
+ENV CXXFLAGS --target=wasm32-wasi --sysroot=/wasi-sysroot -fno-exceptions
```
sbc100, line 53: Given that we don't currently automatically set `-fno-exceptions` in `wasi-sdk.cmake`, does it make sense to set it here?  Seems odd to diverge from the default.

sbc100 (approved):

docker/Dockerfile
anuraaga: Ah thanks - I've found the flag helpful when building with wasi-sdk but agree both should be changed at the same time so removed it for now

sbc100 (approved):

anuraaga:
Thanks for merging @sbc100! I see the action logs look fine and show a pushed image

https://github.com/WebAssembly/wasi-sdk/actions/runs/3787010175/jobs/6438424135

But the package isn't showing up here

https://github.com/orgs/WebAssembly/packages?repo_name=wasi-sdk

I believe this could be because of a repository or organization setting preventing public packages - would you be able to check on that? Thanks!

sbc100:
> Thanks for merging @sbc100! I see the action logs look fine and show a pushed image
> 
> https://github.com/WebAssembly/wasi-sdk/actions/runs/3787010175/jobs/6438424135
> 
> But the package isn't showing up here
> 
> https://github.com/orgs/WebAssembly/packages?repo_name=wasi-sdk
> 
> I believe this could be because of a repository or organization setting preventing public packages - would you be able to check on that? Thanks!

Indeed, it shows up for me there, but is marked as "private".  I'll see if I can find the setting to make it public.

sbc100:
It looks like changing the package visibility settings is an organization-wide permission which I don't have.  @dschuff @sunfishcode might be able to do it: https://github.com/orgs/WebAssembly/packages/container/wasi-sdk/settings

abrown:
@anuraaga, once the package permissions are figured out, can you follow this up with a quick note (here or in the project `README`) on how to use this GitHub-hosted image? I'm familiar with Docker but not with the ghcr.io registry or how this image can be used. I would imagine this to look something like `docker run ghcr.io:<something?>`? And would we need to append `clang` on the end of that, e.g.?

anuraaga:
Thanks @abrown! Sent #282 with a README section, there isn't anything special about ghcr.io except the image name happens to start with `ghcr.io/`, all other Docker knowledge will be usable as-is :)

PR #273, opened 2022-12-24
Use older github runners for better compatibilities

yamt:
See the commit messages for relevant errors.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         os:
           - ubuntu-20.04
-          - macos-latest
+          - macos-11
     steps:
       - uses: actions/cache@v3
         with:
@@ -34,10 +34,10 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ~/Library/Caches/ccache
-          key: 0-cache-macos-latest-${{ github.run_id }}
+          key: 0-cache-${{ matrix.os }}-${{ github.run_id }}
           restore-keys: |
-            0-cache-macos-latest
-        if: matrix.os == 'macos-latest'
+            0-cache-${{ matrix.os }}
+        if: startsWith(matrix.os, 'macos-')
       - name: Install wasmtime for tests
         run: curl -f -L --retry 5 https://wasmtime.dev/install.sh | bash -s -- --version v2.0.2
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


TerrorJack:
For linux: to maximize compatability with older glibc versions, we already have a docker build job that performs the build in an ubuntu bionic container, that artifact should be used when the glibc version error pops up.

For macos: I recommend sticking to `macos-latest` and explicitly adding `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0` to the `Makefile` when building LLVM. This ensures we can benefit from future system/toolchain updates while retaining some degree of compatability with older versions of macos.

yamt:
> For linux: to maximize compatability with older glibc versions, we already have a docker build job that performs the build in an ubuntu bionic container, that artifact should be used when the glibc version error pops up.

will it be a part of assets for the next release?
i'm not entirely sure about the release process.

> 
> For macos: I recommend sticking to `macos-latest` and explicitly adding `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0` to the `Makefile` when building LLVM. This ensures we can benefit from future system/toolchain updates while retaining some degree of compatability with older versions of macos.

while it's reasonable to specify the target version, i feel it's simpler to use a fixed version on the ci.

yamt:
> > For linux: to maximize compatability with older glibc versions, we already have a docker build job that performs the build in an ubuntu bionic container, that artifact should be used when the glibc version error pops up.
> 
> will it be a part of assets for the next release? i'm not entirely sure about the release process.

https://github.com/WebAssembly/wasi-sdk/issues/284

yamt:
> Staying with `macos-latest` allows us to benefit from host toolchain updates.

only when github happens to choose 12.
if you want to use the latest, it's better to use `macos-12` than `macos-latest`.

TerrorJack:
> if you want to use the latest, it's better to use macos-12 than macos-latest.

`macos-latest` will default to `macos-12` per https://github.com/actions/runner-images/issues/6384. The point is not pinning to a fixed version, but `latest` will be future proof and do not require manual intervention for future updates.

yamt:
> > if you want to use the latest, it's better to use macos-12 than macos-latest.
> 
> `macos-latest` will default to `macos-12` per [actions/runner-images#6384](https://github.com/actions/runner-images/issues/6384). The point is not pinning to a fixed version, but `latest` will be future proof and do not require manual intervention for future updates.

ok. i understand your point. i don't agree it's a good idea though.

sbc100:
I think the `-DCMAKE_OSX_DEPLOYMENT_TARGET=11.0` solution make sense.  The purpose of that option is to produce backwards compatible binaries which is exactly what we are wanting to do here.

I don't feel strongly about pinning the version of macos used to do the build, but I don't think we should conflate that with the version we are targeting (that is what MAKE_OSX_DEPLOYMENT_TARGET is for).

yamt:
while i still think it's better to use fixed versions of runners, i'm not too unhappy with the current solutions. (CMAKE_OSX_DEPLOYMENT_TARGET and docker)
let's close this.

Issue #272, opened 2022-12-24
cmake toolchain file for wasm32-wasi-pthread


yamt:
currently `wasi-sdk.cmake` has `set(triple wasm32-wasi)`.
there needs a way to use `set(triple wasm32-wasi-pthread)` for threaded apps.

also, `set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")` is required to pass the cmake compiler test (`CMakeTestCCompiler.cmake`) for `wasm32-wasi-pthread`.

a straightforward solution is to have a separate cmake toolchain file, say `wasi-sdk-pthread.cmake` for threaded apps.

abrown:
I think is closed by #274.

PR #274, opened 2022-12-26
add experimenal wasm32-wasi-threads support

yamt:
possible blockers / known issues:

* ~~tls issue https://github.com/WebAssembly/wasi-libc/pull/372~~
* memory import vs export https://github.com/WebAssembly/wasi-threads/issues/22
* `thread-spawn` ABI https://github.com/WebAssembly/wasi-threads/pull/26 https://github.com/WebAssembly/wasi-libc/pull/385

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,6 +196,7 @@ build/config.BUILT:
 	cp src/config/config.sub src/config/config.guess $(BUILD_PREFIX)/share/misc
 	mkdir -p $(BUILD_PREFIX)/share/cmake
 	cp wasi-sdk.cmake $(BUILD_PREFIX)/share/cmake
+	cp wasi-sdk-pthread.cmake $(BUILD_PREFIX)/share/cmake
 	touch build/config.BUILT
 
 build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libcxx.BUILT build/config.BUILT
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,12 @@ build/wasi-libc.BUILT: build/llvm.BUILT
 		AR=$(BUILD_PREFIX)/bin/llvm-ar \
 		NM=$(BUILD_PREFIX)/bin/llvm-nm \
 		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot
+	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
+		CC=$(BUILD_PREFIX)/bin/clang \
+		AR=$(BUILD_PREFIX)/bin/llvm-ar \
+		NM=$(BUILD_PREFIX)/bin/llvm-nm \
+		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
+		THREAD_MODEL=posix
 	touch build/wasi-libc.BUILT
 
 build/compiler-rt.BUILT: build/llvm.BUILT build/wasi-libc.BUILT
--- a/wasi-sdk-pthread.cmake
+++ b/wasi-sdk-pthread.cmake
@@ -1,0 +1,31 @@
+# Cmake toolchain description file for the Makefile
+
+# This is arbitrary, AFAIK, for now.
+cmake_minimum_required(VERSION 3.4.0)
+
+set(CMAKE_SYSTEM_NAME WASI)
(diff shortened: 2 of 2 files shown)


yamt:
https://github.com/WebAssembly/wasi-sdk/issues/272

yamt:
with this and https://github.com/WebAssembly/wasi-sdk/pull/275,
i was able to build ffmpeg with pthread.
https://github.com/yamt/FFmpeg-WASI
https://github.com/yamt/FFmpeg-WASI/releases/tag/yamt-20221229

i was able to use more than 1 cpus with the above binary.
eg.
```
~/git/toywasm/b/toywasm --wasi --wasi-dir videos -- ffmpeg.wasm -i videos/video-15s.avi -c:v libx264 -threads 16 videos/out.mp4
```

yamt:
depending on https://github.com/WebAssembly/wasi-threads/issues/22 we might need to wait for a llvm release with https://reviews.llvm.org/D135898

abrown:
I wouldn't worry too much about using your own fork here; even if wasi-libc doesn't have the TLS PR merged yet this would still be a step in the right direction. We just would want to make sure that any missing pieces like that PR get updated in the submodule before @sunfishcode publishes a new wasi-sdk release, which I hear might be in the next few weeks (?).

yamt:
> I wouldn't worry too much about using your own fork here; even if wasi-libc doesn't have the TLS PR merged yet this would still be a step in the right direction. We just would want to make sure that any missing pieces like that PR get updated in the submodule before @sunfishcode publishes a new wasi-sdk release, which I hear might be in the next few weeks (?).

ok. i will drop the "use my fork" commit. instead, i mentioned know issues in the description of this PR.

sunfishcode (approved):
Looks good to me.

abrown:
@sunfishcode, you'll have to merge this (I don't have permissions here but I'm a +1 for this PR). @yamt, do we still need to update the wasi-libc commit in the submodule? If so, that can probably be its own PR.

yamt:
> @sunfishcode, you'll have to merge this (I don't have permissions here but I'm a +1 for this PR). @yamt, do we still need to update the wasi-libc commit in the submodule? If so, that can probably be its own PR.

the latest wasi-libc should be ok.
i haven't checked if the submodule is latest enough.

my concern at this point is the coming ABI change from: https://github.com/WebAssembly/wasi-threads/pull/26
cf. https://github.com/WebAssembly/wasi-libc/pull/387

sbc100:
> How stable we consider wasi-sdk? Are we ok at that stage to introduce backward incompatible changes (that can be detected at compile time)?

Are you asking in general, or specifically regarding this PR?  This PR doesn't break any existing thing right, it add a completely new set of target libraries while leaving the rest of the SDK untouched?

In the general case, WASI is not yet stable or standardized, and neither is wasi-sdk.  Of course, we can and should always do our best to avoid breaking existing code, but we are not at the stage in development of WASI where we rule out such things.  Even the core components on which we build wasi-sdk (llvm and musl) make breaking changes from time to time so it not practical to rule out all breaking changes.

loganek:
> Are you asking in general, or specifically regarding this PR? 

Mainly asked in the context of @yamt  's concerns about ABI changes. Looks like wam32-wasi-threads target is even more experimental than the rest of the wasi-sdk, so I don't think we have to wait for the ABI to be finalized?

abrown:
Right, to echo @sbc100, we cannot make too strong guarantees about about the wasi-threads ABI. We have an ABI that is good enough for users to experiment with and we can accompany the prerelease of all of this with some warnings in the documentation.

loganek (approved):
LGTM

abrown (approved):

TerrorJack:
There doesn't seem to be libcxx/libcxxabi builds for `wasm32-wasi-threads` yet, is it an oversight?

abrown:
No, just not enough time to do everything. Are you thinking of contributing something like this in https://github.com/WebAssembly/wasi-libc/issues/392?

whitequark:
> There doesn't seem to be libcxx/libcxxabi builds for `wasm32-wasi-threads` yet, is it an oversight?

https://github.com/WebAssembly/wasi-sdk/pull/301

denis-belov:
Hello guys! Does it mean that `malloc()` will be thread-safe from now?

PR #275, opened 2022-12-29
Only strip files under the bin directory

yamt:
Stripping other files (eg. *.a) doesn't seem the original intention and it sometimes has ill effects like breaking archive TOC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,7 +199,7 @@ build/config.BUILT:
 build: build/llvm.BUILT build/wasi-libc.BUILT build/compiler-rt.BUILT build/libcxx.BUILT build/config.BUILT
 
 strip: build/llvm.BUILT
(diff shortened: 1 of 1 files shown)


sbc100 (approved):

TerrorJack:
Another possible approach would be adding `-DCMAKE_EXE_LINKER_FLAGS="-s"` to the LLVM configure step, so the linker produces stripped executables directly and we can get rid of the `strip` logic altogether.

sbc100:
> Another possible approach would be adding `-DCMAKE_EXE_LINKER_FLAGS="-s"` to the LLVM configure step, so the linker produces stripped executables directly and we can get rid of the `strip` logic altogether.

Performing the stripped explicitly after the build feels more robust/obvious and makes sense to me from my old debian maintainer days.

TerrorJack (approved):

PR #144, opened 2020-06-14
Run tests under wasmtime as part of the CI

sbc100:

PR #278, opened 2023-01-03
Run tests under wasmtime as part of the CI

TerrorJack:
With approval from @sbc100, this is a rework of #144. It's checked in x64 linux/macos pipelines. Also contains minor fixes in the expected files.

We use `wasmtime-2.0.2` since it suffices for the time being (no fancy wasm extensions in our sysroot for now), and the error messages in later versions of `wasmtime` contain indeterministic addresses which makes comparison harder.


sbc100 (approved):

TerrorJack:
Pipeline just passed :)

PR #279, opened 2023-01-03
Avoid unnecessary git checkout overhead

TerrorJack:
The git checkout step in the current CI pipelines are pretty slow. The overhead mainly comes from the `llvm-project` submodule, however, we only need a shallow checkout of it, and full checkout with all tags is only needed for `wasi-sdk` parent repo. This PR shall further speed up all CI pipelines.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,9 +40,15 @@ jobs:
         if: matrix.os == 'macos-latest'
       - name: Install wasmtime for tests
         run: curl -f -L --retry 5 https://wasmtime.dev/install.sh | bash -s -- --version v2.0.2
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
         with:
-          submodules: true
+          fetch-depth: 0
+      # We can't use `--depth 1` here sadly because the GNU config
+      # submodule is not pinned to a particular tag/branch. Please
+      # bump depth (or even better, the submodule), in case of "error:
+      # Server does not allow request for unadvertised object" in the
+      # future.
+      - run: git submodule update --init --depth 16 --jobs 3
       - name: Install ccache, ninja (macOS)
         run: brew install ccache ninja
         if: matrix.os == 'macos-latest'
@@ -96,7 +102,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          submodules: true
+      - run: git submodule update --init --depth 16 --jobs 3
       - name: Build
         shell: msys2 {0}
         run: |
@@ -125,9 +131,11 @@ jobs:
           restore-keys: |
             0-cache-ubuntu-bionic
 
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
         with:
-          submodules: true
+          fetch-depth: 0
+
+      - run: git submodule update --init --depth 16 --jobs 3
 
       - uses: docker/login-action@v2
         with:


TerrorJack:
The speed up at git checkout step, compared to latest `main` pipeline:

<table>
<tr>
	<td>Job
	<td>Before
	<td>After
<tr>
	<td>ubuntu-latest
	<td>5m40s
	<td>1m49s
<tr>
	<td>macos-latest
	<td>5m33s
	<td>2m19s
<tr>
	<td>windows x64
	<td>7m15s
	<td>2m33s
<tr>
	<td>windows x86
	<td>8m8s
	<td>3m13s
<tr>
	<td>docker build
	<td>5m7s
	<td>1m54s
</table>

sbc100 (approved):
Why not just use depth 1?

TerrorJack:
> Why not just use depth 1?

A mere `--depth 1` will make `git` fail to checkout `config`, since its current revision is not the tip of any git ref.

sbc100:
> > Why not just use depth 1?
> 
> A mere `--depth 1` will make `git` fail to checkout `config`, since its current revision is not the tip of any git ref.

So its not on a git ref but it happens to be within 16 commits of one?   That seems worth a comment.   I also wonder if it worth bumping to a tagged release so avoid this?

TerrorJack:
> That seems worth a comment.

Sure, will do.

> I also wonder if it worth bumping to a tagged release so avoid this?

https://git.savannah.gnu.org/cgit/config.git/refs/ doesn't seem to provide any tagged releases!

.github/workflows/main.yml
```diff
@@ -40,9 +40,12 @@ jobs:
         if: matrix.os == 'macos-latest'
       - name: Install wasmtime for tests
         run: curl -f -L --retry 5 https://wasmtime.dev/install.sh | bash -s -- --version v2.0.2
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
         with:
-          submodules: true
+          fetch-depth: 0
+      # Bump depth (or even better, submodule), in case of "error:
+      # Server does not allow request for unadvertised object"
```
sbc100, line 47: How about:

```
# We can't use `--depth 1` here sadly because the GNU config submodule is not pinned to a particular tag/branch. 
```

I do worry that this hardcoded value of 16 could become out-of-date as the config.git repository receives more commits, but maybe we just deal with that when it happens.

TerrorJack: Done, thanks for the addition. I've also bumped `config.git`, and this repo updates rather slowly, so we're pretty unlikely to bump into this issue for a very long time.

.github/workflows/main.yml
```diff
@@ -40,9 +40,15 @@ jobs:
...
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
         with:
-          submodules: true
+          fetch-depth: 0
+      # We can't use `--depth 1` here sadly because the GNU config
+      # submodule is not pinned to a particular tag/branch. Please
+      # bump depth (or even better, the submodule), in case of "error:
+      # Server does not allow request for unadvertised object" in the
+      # future.
```
sbc100, line 50: I guess if/when https://github.com/actions/checkout/issues/367 is fixed we could possibly revert this?

sbc100 (approved):

.github/workflows/main.yml
TerrorJack: Yes, by then we can use a single `actions/checkout` step to handle submodules as well.

Issue #284, opened 2023-01-10
Release process should be documented


yamt:
or even automated.

at least it should be clear where release asserts came from.

cf. https://github.com/WebAssembly/wasi-libc/issues/326#issuecomment-1362055334

thewtex:
Heads up, the assets are currently missing from wasi-sdk-18: https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-18

sunfishcode:
That's because I haven't done the manual steps yet :-}. Also, there's a regression being discussed in https://github.com/WebAssembly/wasi-libc/pull/377, which only affects users using -Wl,--initial-memory, and there's a possibility we may do another release soon to fix it.

yamt:
> manually download the artifacts from the latest CI build

@TerrorJack suggested to use linux artifacts from docker job: https://github.com/WebAssembly/wasi-sdk/pull/273#issuecomment-1373879491
is it what you do?

yamt:
> I download the artifacts from a CI run, such as here: https://github.com/WebAssembly/wasi-sdk/actions/runs/3889687857#artifacts

do you mean to include binary from both of dist-ubuntu-bionic and dist-ubuntu-latest?

sunfishcode:
I look at the artifacts from the previous release and try to upload the same artifacts.

TerrorJack:
@sunfishcode It seems the `dist-ubuntu-latest` artifact has been the choice of the linux binary release for the past few releases. It may also be good to also upload `dist-ubuntu-bionic` and name it something like ` wasi-sdk-19.0-ubuntu-bionic.tar.gz `, and mention it in the release notes, so when people want to use it in a distro with outdated glibc, they can choose it.

PR #287, opened 2023-01-11
Add the `wasm32-wasi-pthread` output to the sysroot

abrown:
Now that the `wasm32-wasi-pthread` target in wasi-libc has been worked on for some amount of time, this change allows it to be released as a part of wasi-sdk packages. It should be noted somewhere (a compiler warning? README?) that this new target is not entirely stable, but that can be added as a follow-on commit. This change is mainly focused on ensuring that the build artifacts contain the new objects.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,12 @@ build/wasi-libc.BUILT: build/llvm.BUILT
 		AR=$(BUILD_PREFIX)/bin/llvm-ar \
 		NM=$(BUILD_PREFIX)/bin/llvm-nm \
 		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot
+	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
+		CC=$(BUILD_PREFIX)/bin/clang \
+		AR=$(BUILD_PREFIX)/bin/llvm-ar \
+		NM=$(BUILD_PREFIX)/bin/llvm-nm \
+		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
+		THREAD_MODEL=posix
 	touch build/wasi-libc.BUILT
 
 build/compiler-rt.BUILT: build/llvm.BUILT build/wasi-libc.BUILT


abrown:
One note on this: I think we may want to change the name of this target to `wasm32-wasi-threads` in a future commit to wasi-libc and pull that in to wasi-sdk. The reason is that I think we should try to avoid confusion if at all possible and over in rustc (https://github.com/rust-lang/compiler-team/issues/574) it is more likely that `wasm32-wasi-threads` is selected as the target name, since pthreads are not relevant to the Rust standard library.

yamt:
see https://github.com/WebAssembly/wasi-sdk/pull/274

sbc100:
> One note on this: I think we may want to change the name of this target to `wasm32-wasi-threads` in a future commit to wasi-libc and pull that in to wasi-sdk. The reason is that I think we should try to avoid confusion if at all possible and over in rustc ([rust-lang/compiler-team#574](https://github.com/rust-lang/compiler-team/issues/574)) it is more likely that `wasm32-wasi-threads` is selected as the target name, since pthreads are not relevant to the Rust standard library.

Are you sure pthreads is not relevant to the Rust standard library?   I feel like I've asked this question before but doesn't the rust standard library want to use pthreads under the hood?  Doesn't it do this already on platforms like linux and mac?   Wouldn't it make it hard to link in other C libraries that are built on pthreads if rust doesn't use pthreads but instead rolled its own threads on top of wasi-thread-spawn (or equivalently if rust use close directly instead of pthread_create on linux?)

abrown:
> Are you sure pthreads is not relevant to the Rust standard library? I feel like I've asked this question before but doesn't the rust standard library want to use pthreads under the hood?

Oh, I'm not saying it shouldn't and in fact I think I've discussed with a few people that we might want to do just that--use wasi-libc's pthread support. What I'm trying to say is that "pthread" as a name could confuse things; whether or not the Rust stdlib uses pthreads is kind of an internal detail that we probably don't want users getting hung up on, IMO. It seems preferable that users have 1) a consistent target and 2) a target name that reminds us of the Wasm threads proposal and the wasi-threads proposal, not one use of it. (Hopefully I explained this well over in https://github.com/WebAssembly/wasi-libc/pull/381).

sbc100 (approved):

yamt:
https://github.com/WebAssembly/wasi-sdk/pull/274 has been containing the identical change and it's more complete IMO.
it has been a draft because i'm still waiting for a critical fix in wasi-libc: https://github.com/WebAssembly/wasi-libc/pull/372

abrown:
I agree that #274 is more complete. I think I will close this one; @yamt, can you modify #284 to use the new target name that just merged, `wasm32-wasi-threads`?

PR #295, opened 2023-02-11
ci: document and script the release process

abrown:
Previously, the release process for wasi-sdk was undocumented and manual. The `RELEASING.md` document in this commit describes the steps necessary to publish a new version of `wasi-sdk` as a GitHub release and provides helpful scripts to automate some of the steps.

With this in place, a future change could add a workflow trigger that allows running the steps automatically in GitHub actions. Keeping the steps as scripts in the repository, however, allows developers to re-run steps themselves in the (likely) case that something goes awry.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -7,7 +7,7 @@ To publish a new version of `wasi-sdk` as a GitHub release:
    package version (use `git show wasi-sdk-...` to show other tag messages).
 
    ```shell script
-   export TAG=wasi-sdk-1
+   TAG=wasi-sdk-1
    git tag -a $TAG
    git push origin $TAG
    ```
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,49 @@
+# Release Process
+
+To publish a new version of `wasi-sdk` as a GitHub release:
+
+1. Tag a commit with an annotated tag. Note that this must be an annotated tag,
+   not a lightweight tag, so that `version.sh` can use it for calculating the
+   package version (use `git show wasi-sdk-...` to show other tag messages).
+
+   ```shell script
+   export TAG=wasi-sdk-1
+   git tag -a $TAG
+   git push origin $TAG
+   ```
+
+2. Find a successful workflow that CI has run for the tag. That successful
+   workflow run will have build artifacts that need to be attached to the
+   release. One could search around in the GitHub [actions], but the following
+   script will list completed workflows for a tag (get a token [here][tokens]):
+
+   ```shell script
+   ci/get-workflows-for-tag.sh $TAG $GITHUB_TOKEN
+   ```
+
+   [actions]: https://github.com/WebAssembly/wasi-sdk/actions
+   [tokens]: https://github.com/settings/tokens
+
+3. Download and unzip the workflow artifacts. Note that artifacts with `+m` or
+   `.m` suffixes indicate that the Git tree was modified. Expect some duplicates
+   since some of the same artifacts are built on multiple CI runners (e.g.,
+   Windows, MacOS, Linux). The following script does all of this automatically:
+
+   ```shell script
+   ci/download-workflow-artifacts.sh $TAG $WORKFLOW_RUN_ID $GITHUB_TOKEN
+   ```
+
+4. Draft a new release. This could be done [manually][releases] but the
+   following script simplifies the uploading of all the files and auto-generates
+   the release description:
+
+  ```shell script
+  ci/draft-release.sh $TAG $ARTIFACTS_DIR $GITHUB_TOKEN
+  ```
+
+  [releases]: https://github.com/WebAssembly/wasi-sdk/releases
+
+5. Publish the release; the previous step only creates a draft. Follow the link
+   in the previous step or navigate to the GitHub [releases] to review the
+   description, commit, tag, and assets before clicking "Publish"
+
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -46,4 +46,3 @@ To publish a new version of `wasi-sdk` as a GitHub release:
 5. Publish the release; the previous step only creates a draft. Follow the link
    in the previous step or navigate to the GitHub [releases] to review the
    description, commit, tag, and assets before clicking "Publish"
-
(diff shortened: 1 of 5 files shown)


RELEASING.md
```diff
@@ -0,0 +1,49 @@
+# Release Process
+
+To publish a new version of `wasi-sdk` as a GitHub release:
+
+1. Tag a commit with an annotated tag. Note that this must be an annotated tag,
+   not a lightweight tag, so that `version.sh` can use it for calculating the
+   package version (use `git show wasi-sdk-...` to show other tag messages).
+
+   ```shell script
+   export TAG=wasi-sdk-1
```
sbc100, line 10: Remove `export` here.

RELEASING.md
```diff
@@ -0,0 +1,49 @@
...
+  ```shell script
+  ci/draft-release.sh $TAG $ARTIFACTS_DIR $GITHUB_TOKEN
+  ```
+
+  [releases]: https://github.com/WebAssembly/wasi-sdk/releases
+
+5. Publish the release; the previous step only creates a draft. Follow the link
+   in the previous step or navigate to the GitHub [releases] to review the
+   description, commit, tag, and assets before clicking "Publish"
+
```
sbc100, line 49: Remove extra newline here?

ci/draft-release.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
```
sbc100, line 1: Did you write all these new shell scripts from scratch?

Do you expect these scripts to run during ci?  If not perhaps they should just live under `scripts/` or at the top level like the rest of the current `.sh` scripts?

ci/draft-release.sh
```diff
@@ -0,0 +1,71 @@
...
+        -X POST \
+        -H "Accept: application/vnd.github+json" \
+        -H "Authorization: Bearer ${GITHUB_TOKEN}"\
+        -H "X-GitHub-Api-Version: ${GITHUB_API_VERSION}" \
+        -H "Content-Type: application/octet-stream" \
+        "${UPLOAD_URL}?name=${FILE}" \
+        --data-binary "@${FROM}")
+done
+
+>&2 echo
```
sbc100, line 68: TIL you can put the redirect before the command.. how strange.

sbc100 (commented):
Wow, nice work.   I didn't get a chance to review those new scripts but this looks like a create step forward.

ci/download-workflow-artifacts.sh
```diff
@@ -0,0 +1,102 @@
...
+
+# Unzip the workflow artifacts into a `release` directory.
+pushd $TMP_DIR > /dev/null
+mkdir release
+ls -1 *.zip | xargs -n1 unzip -q -o -d release
+# Some explanation:
+#   -1 prints each file on a separate line
+#   -n1 runs the command once for each item
+#   -q means quietly
+#   -o allows unzip to overwrite existing files (e.g., multiple copies of `libclang_rt.builtins-wasm32-wasi-...`)
```
yamt, line 94: for linux, assets from the docker job should be used.
btw, please fix the linux assets for wasi-sdk-20+threads.

ci/download-workflow-artifacts.sh
```diff
@@ -0,0 +1,102 @@
...
+#   retrieve the artifacts from
+# - a GitHub access token, see https://github.com/settings/tokens
+#
+# Usage: download-workflow-artifacts.sh <release tag> <workflow run ID> <token>
+
+TAG=$1
+WORKFLOW_RUN_ID=$2
+GITHUB_TOKEN=$3
+GITHUB_API_VERSION=2022-11-28
+GITHUB_API_URL=https://api.github.com/repos/WebAssembly/wasi-sdk
```
yamt, line 19: why do you prefer to have a low-level script here while there are many handy commands/scripts (eg. `hub`) and even github actions available?

ci/download-workflow-artifacts.sh
abrown: I'm not sure I understand why?

ci/download-workflow-artifacts.sh
abrown: Not everyone has all of that installed; I went with an environment I thought @sunfishcode would like 😁. (The idea was that these scripts would be runnable both from CI or from a maintainer's shell if something goes wrong).

ci/draft-release.sh
abrown: Heh, yeah, the explanation on StackOverflow was that this makes it more obvious that we are redirecting to stderr and I bought the argument.

ci/draft-release.sh
abrown: Yes, I wrote them as I fought through issues with CI/Git/GitHub while trying to publish the wasi-sdk-20+threads pre-release (it was a fight: not every CI run has the right tag applied, e.g.). I had to do things enough times that this work already paid for itself, but in the back of my head it seemed like a good idea to record how one might publish new wasi-sdk releases. I was hoping at some point to integrate these in CI with a GitHub workflow trigger but that could be done later. I don't mind moving them to `scripts` or the top-level directory... tell me which is best and I'll move them. (Maybe we move all the `*.sh` files to `scripts`?).

ci/download-workflow-artifacts.sh
yamt: assets from ubuntu job (dist-ubuntu-latest) requires newer glibc.
assets from docker job (dist-ubuntu-bionic) is compatible with wider distributions.

ci/download-workflow-artifacts.sh
yamt: i suspect it's easier to install one of them than maintaining this script by ourselves. but i don't insist.

ci/draft-release.sh
abrown: I talked to @sunfishcode a bit more about this kind of PR and I think I'm going to go ahead and merge soon; @sbc100, any strong preference on where to put these scripts?

sbc100: Not really no.

PR #297, opened 2023-02-15
wasi-sdk-pthread.cmake: add --import-memory

yamt:


TerrorJack:
Shall we also add `--shared-memory` and a large `--max-memory` default value?

yamt:
> Shall we also add `--shared-memory` and a large `--max-memory` default value?

`--shared-memory` is automatically added for -pthead.

i don't think there is a good single value for `--max-memory`. it depends on apps.

yamt:
i marked this a draft because this needs llvm >=16.

abrown:
I am also wondering if we should rename this file `wasi-sdk-threads.cmake`?

yamt:
> I am also wondering if we should rename this file `wasi-sdk-threads.cmake`?

i don't think so as what this provides is pthread.

yamt:
this is ready as we now ship llvm 16.

wasi-sdk-pthread.cmake
```diff
@@ -9,6 +9,11 @@ set(CMAKE_SYSTEM_PROCESSOR wasm32)
 set(triple wasm32-wasi-threads)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
+# wasi-threads requries --import-memory.
```
abrown, line 12:
```suggestion
# wasi-threads requires --import-memory.
```

abrown (approved):

wasi-sdk-pthread.cmake
yamt: fixed. thank you.

PR #296, opened 2023-02-14
Pre-release `wasi-sdk-20+threads`

abrown:
__DO NOT MERGE!__

This PR is meant as a visual reminder of how the [`wasi-sdk-20+threads`](https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-20%2Bthreads) pre-release was created. This branch includes changes to:
 - update the LLVM source code to v16.0.0-rc2
 - update the wasi-libc source code to its latest HEAD
 - fix an issue with the major version used in the wasi-sdk path
 - and handle some additional kinds of string (e.g., `+threads`) in the wasi-sdk version tags

Of these, the last change (the additional `version.sh` strings) can certainly be dropped for a future release of `wasi-sdk-20` and the source code bumps likely need to be to even later commits. This PR may be useful, however, as a record of what was was done for `wasi-sdk-20+threads`.

 cc: @sunfishcode, @sbc100

PR #301, opened 2023-02-23
Build libcxxabi and libcxx with threads enabled

whitequark:
Depends on https://github.com/WebAssembly/wasi-libc/pull/397.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -153,8 +153,8 @@ LIBCXX_CMAKE_FLAGS = \
     -DLLVM_CONFIG_PATH=$(ROOT_DIR)/build/llvm/bin/llvm-config \
     -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
     -DCXX_SUPPORTS_CXX11=ON \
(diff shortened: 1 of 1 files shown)


whitequark:
I've built [YoWASP](https://yowasp.org) [nextpnr](https://github.com/YosysHQ/nextpnr) with parallel refinement enabled and I can report that the improvements running on wasmtime are very similar to those seen with a native build.

abrown:
@whitequark, thanks for this PR! I am fine with this but I feel like @sunfishcode should take a look: do we want to merge this in `main`? I think it makes sense to do so, but it is related to the decision on whether the `wasi-sdk-20` &mdash; the real release, not my pre-release &mdash; contains all of the `wasm32-wasi-threads` target stuff. I would assume yes, but wanted to leave it up to you. (Once this is merged, I could patch it into https://github.com/WebAssembly/wasi-sdk/pull/296 and rebuild the `wasi-sdk-20+threads` pre-release with this to try things out).

sunfishcode:
Looks good to me!

Issue #303, opened 2023-03-10
Using wasi-sdk-20 pre-release on Ubuntu 20.04


eloparco:
Hello,
In a GitHub action job on Ubuntu 20.04, I'm trying to use the wasi-sdk-20 pre-release, but I'm getting the following eror:
```bash
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /opt/wasi-sdk/bin/clang)
/opt/wasi-sdk/bin/clang: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /opt/wasi-sdk/bin/clang)
```
Is there any planned fix to support older versions of glibc?
Thanks

abrown:
@yamt had mentioned this in a comment previously, but I personally was not too interested in maintaining compatibility with a bunch of old glibc versions ("how far to go back?" "would this cause friction with future integration?" etc.). IIUC, this boils down to using Ubuntu's 20.04 image during compilation instead of 22.04; I don't understand what restrictions are preventing other projects from also using 22.04? Is this request of wasi-sdk just a symptom of another issue somewhere else? That being said, I'm actually not opposed to supporting some older versions of glibc, though; I would gladly review PRs along these lines if we figure out a limit to the versions supported and it meshes with the work I did to script the release process in https://github.com/WebAssembly/wasi-sdk/pull/295.

yamt:
> @yamt had mentioned this in a comment previously, but I personally was not too interested in maintaining compatibility with a bunch of old glibc versions ("how far to go back?" "would this cause friction with future integration?" etc.). IIUC, this boils down to using Ubuntu's 20.04 image during compilation instead of 22.04; I don't understand what restrictions are preventing other projects from also using 22.04? Is this request of wasi-sdk just a symptom of another issue somewhere else? That being said, I'm actually not opposed to supporting some older versions of glibc, though; I would gladly review PRs along these lines if we figure out a limit to the versions supported and it meshes with the work I did to script the release process in #295.

as far as i know, no one is asking "compatibility with a bunch of old glibc versions".
just a compatibility with commonly used distributions like ubuntu 20.04.
[...]

PR #307, opened 2023-03-15
download-workflow-artifacts.sh: skip dist-ubuntu-latest

yamt:
Exclude dist-ubuntu-latest to prefer dist-ubuntu-bionic, which is compatible with wider distributions.
cf.
https://github.com/WebAssembly/wasi-sdk/pull/273#issuecomment-1373879491
https://github.com/WebAssembly/wasi-sdk/issues/303


abrown (approved):
I think this makes sense, thanks! Should we test this out with the `wasi-sdk-20+threads` pre-release or just wait until `wasi-sdk-20`?

PR #314, opened 2023-03-20
llvm: update to LLVM 16.0.0

abrown:
This builds on #313 (merge that first) to update the LLVM sources to point to the [release] version of LLVM 16.0.0. It also contains a Makefile fix for a path [change] for `libclang_rt.builtins-wasm32.a`.

[release]: https://github.com/llvm/llvm-project/releases/tag/llvmorg-16.0.0
[change]: https://reviews.llvm.org/D125860


Makefile
```diff
@@ -38,6 +38,7 @@ BASH=
 endif
 
 CLANG_VERSION=$(shell $(BASH) ./llvm_version.sh $(LLVM_PROJ_DIR))
+CLANG_MAJOR_VERSION=$(shell $(BASH) ./llvm_version_major.sh $(LLVM_PROJ_DIR))
```
sbc100, line 41: I don't think we need both of these.. can you just modify the existing llvm_version.sh to report just the major version?'

Otherwise LGTM

abrown: Take a look at that last commit: the "make sure this isn't version 9 or 10" code also needed to change. (I wonder if there is an even better way to do that?)

PR #316, opened 2023-03-24
Update docker build to use LLVM 16

anuraaga:
As LLVM 16 was released, it seems like a good idea to update. Notably, this will allow the docker build to provide a threads-enabled build as well - I may be missing it but AFAICT, threads hasn't been wired into GitHub CI but when it is it would be easy to publish the docker image too


docker/Dockerfile
```diff
@@ -33,11 +34,7 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 COPY --from=dist /wasi-sdk/share/wasi-sysroot/ /wasi-sysroot/
-# The path to the rt directory contains the LLVM patch version which is not reflected in the LLVM apt repository
-# or package. To make adding the RT robust to changing patch versions without needing to duplicate the folder
-# content, we symlink after extracting using a bash glob to resolve the patch version
-ADD dist/libclang_rt.builtins-wasm32-wasi-*.*.tar.gz /wasi-sysroot-clang_rt
-RUN ln -s /wasi-sysroot-clang_rt/lib/wasi/ $(echo /usr/lib/llvm-${LLVM_VERSION}/lib/clang/${LLVM_VERSION}.*)/lib/wasi
+COPY --from=dist /wasi-sysroot-clang_rt/lib/wasi /usr/lib/llvm-${LLVM_VERSION}/lib/clang/${LLVM_VERSION}/lib/wasi
```
anuraaga, line 37: It seems that the problematic patch version in path for this isn't there anymore

sbc100 (approved):

abrown (approved):

Issue #236, opened 2022-05-18
Add binary release for Linux on Apple Silicon M1


talg:
I think an increasing number of us are running on Linux on Apple Silicon with Parallels in a VM, and support would be great. I can always build from source, but I would guess there will be plenty of others who will run into this.

There is already an issue for this on MacOS #221.

sascha1337:
wen

weiqingtangx:
Guys, do you have a plan for this?

sbc100:
Instructions for building from source are in the README: https://github.com/WebAssembly/wasi-sdk

I don't know of anyone working on a linux arm64 build.  We build our images using github CI I guess it depends on when we can actions runing on arm64 linux.. or somebody could figure out to do a cross-compile on x86?

sascha1337:
followup, needed

talg:
https://github.blog/changelog/2022-08-09-github-actions-self-hosted-runners-now-support-apple-m1-hardware/

gzurl:
+1
In my case, I'm running a container on top of an Apple M1, so I'm missing the `-linux-arm64` flavor.

gzurl:
More info: Building from sources from a linux-arm64 container (tried with `ubuntu:22.04` and bitnami/minideb) fails.

In both containers, the error message is:

```
[...]

PR #321, opened 2023-04-05
Fix build on non-x86_64 Linux

alexcrichton:
It looks like some macOS-specific CMake options to LLVM end up forcing an x86_64 build of some blake3 intrinsics (or something like that) which causes the build to fail on AArch64 Linux, for example. I found, though, that when removing these options it was then possible to build a toolchain for AArch64 Linux.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ BASH=
 
 endif
 
+ifeq ($(shell uname),Darwin)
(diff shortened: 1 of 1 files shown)


sbc100:
Is this the same issue that is coming up in https://github.com/WebAssembly/wasi-sdk/issues/236?   I wonder if we should link to a bug here, or at least a comment.  Seems like a upstream llvm bug in the build system.

alexcrichton:
Yes I believe it's the same issue.

abrown (approved):

PR #322, opened 2023-04-05
Pass through extra flags to libcxx build

alexcrichton:
This commit passes through `EXTRA_CFLAGS` to the libcxx build, along with `EXTRA_CXXFLAGS`. This models after the pattern in `wasi-libc` for passing these flags to thread more flags through.


abrown (approved):
Makes sense to me!

PR #327, opened 2023-05-08
tests: use latest wasmtime for testing

TerrorJack:
This patch enables using latest version of wasmtime for testing. This should also make it possible to run tests for wasm32-wasi-threads in the future.


abrown (approved):
LGTM!

PR #331, opened 2023-06-23
Improve `README.md`

abrown:
This changes the front-page documentation to:
- use out-of-line links in more places
- mention the need for `libclang_rt.builtins-wasm32.a` when using
  standard Clang
- mention how to use a `wasi-libc` sysroot
- explain the experimental status of `wasm32-wasi-threads

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This repository contains no compiler or library code itself; it uses
 git submodules to pull in the upstream Clang and LLVM tree, as well as the
 wasi-libc tree.
 
-The libc portion of this SDK is the [wasi-libc].
+The libc portion of this SDK is maintained in [wasi-libc].
 
 [wasi-libc]: https://github.com/WebAssembly/wasi-libc
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 ## Quick Start
 
-[Download SDK packages here.](https://github.com/WebAssembly/wasi-sdk/releases)
+[Download SDK packages here.][releases]
+
+[releases]: https://github.com/WebAssembly/wasi-sdk/releases
 
 ## About this repository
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,25 +2,31 @@
...
 
 ## About this repository
 
 This repository contains no compiler or library code itself; it uses
 git submodules to pull in the upstream Clang and LLVM tree, as well as the
 wasi-libc tree.
 
-The libc portion of this SDK is the
-[wasi-libc](https://github.com/WebAssembly/wasi-libc).
+The libc portion of this SDK is the [wasi-libc].
```
sbc100, line 15: Drop the second "the"?

How about "lives in" or "is maintained in the separate  [wasi-libc] repository"

sbc100 (approved):
Nice!

README.md
```diff
@@ -2,25 +2,31 @@
...
 convenience.
 
 One could also use a standard Clang installation, build a sysroot from the
-sources mentioned above, and compile with
-"--target=wasm32-wasi --sysroot=/path/to/sysroot".
+sources mentioned above, and compile with `--target=wasm32-wasi
+--sysroot=/path/to/sysroot`. In this scenario, one would also need the
+`libclang_rt.builtins-wasm32.a` objects available separately in the [release
+downloads][releases] which must be extracted into
+`$CLANG_INSTALL_DIR/$CLANG_VERSION/lib/wasi/`.
```
abrown, line 29: Actually, is this true? I know this is true for building wasi-libc (see [building from source](https://github.com/WebAssembly/wasi-libc#building-from-source)) from having to do this myself but is it the case for the average program?

sbc100: I'm not sure exactly what you man.

If your question is "can you get away without `libclang_rt.builtins-wasm32.a`" then the answer is no.  All programs pretty much end up depending on this library and clang adds it to the link line by default.

Also, wasi-libc is not a program (its just a static library) so technically you don't need `libclang_rt.builtins-wasm32.a` to build wasi-libc.. but in order to use wasi-libc in any kind of normal program, that is when you would need the builtins library (i.e. when you actually want to link something).

Or did I misunderstand your question?

sbc100 (approved):

README.md
abrown: Ok, yeah, your response makes sense:

> If your question is "can you get away without libclang_rt.builtins-wasm32.a" then the answer is no

That's what I thought but I started to confuse myself. So the language I'm using here does make sense: we need the builtins to just use vanilla Clang &mdash; it isn't as easy as "just point to a sysroot" as we previously had documented here.

Separately,

> technically you don't need libclang_rt.builtins-wasm32.a to build wasi-libc

But the "we need the builtins" is documented in the wasi-libc instructions; maybe we should discuss that in https://github.com/WebAssembly/wasi-libc/pull/425?

PR #335, opened 2023-07-12
Set CMake module path

arvid-norlander:
This is a redo/update of PR #182, as that was by a different author I cannot update that PR, but have to do a separate one.

Compared to the previous PR, this adds testing of the cmake setup as well.

Fixes: #181

### DIFF
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 set -ueo pipefail
 
-# Top-level test runner. Usage is "run.sh" to run tests in compile-only mode,
-# or "run.sh <runwasi>" where <runwasi> is a WASI-capable runtime to run the
-# tests in full compile and execute mode.
+# Top-level test runner. Usage is "run.sh <path to wasi-sdk>" to run tests
+# in compile-only mode, or "run.sh <path to wasi-sdk> <runwasi>" where
+# <runwasi> is a WASI-capable runtime to run the tests in full compile and
+# execute mode.
 #
 # By default this script will look for `clang` and `clang++` in $PATH and
 # assume that they are correctly configured with the sysroot in the default
@@ -12,10 +13,16 @@ set -ueo pipefail
 #  export CXX="<wasi-sdk>/bin/clang++ --sysroot <wasi-sdk>/share/wasi-sysroot"
 #  export CC="<wasi-sdk>/bin/clang --sysroot <wasi-sdk>/share/wasi-sysroot"
 #
+if [ $# -lt 1 ]; then
+    echo "Path to WASI SDK is required"
+    exit 1
+fi
+
+wasi_sdk="$1"
 
 # Determine the wasm runtime to use, if one is provided.
-if [ $# -gt 0 ]; then
-    runwasi="$1"
+if [ $# -gt 1 ]; then
+    runwasi="$2"
 else
     runwasi=""
 fi
@@ -26,6 +33,7 @@ CXX=${CXX:=clang++}
 
 echo $CC
 echo $CXX
+echo "SDK: $wasi_sdk"
 
 cd $testdir/compile-only
 for options in -O0 -O2 "-O2 -flto"; do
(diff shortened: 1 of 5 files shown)


arvid-norlander:
I seem to have an unrelated failure in:

```
===== Testing with -O0 =====
Testing abort.c...
Testing abort.c...
--- abort.c.stderr.expected     2023-07-07 13:13:40.708014588 +0200
+++ abort.c.-O0.stderr.observed.filtered        2023-07-12 10:19:53.021388399 +0200
@@ -1,6 +1,2 @@
-Error: failed to run main module `abort.c.---.wasm`
-
-Caused by:
-    0: failed to invoke command default
-    1: error while executing at wasm backtrace:
-    2: wasm trap: wasm `unreachable` instruction executed
+error: RuntimeError: unreachable
+╰─▶ 1: RuntimeError: unreachable
make: *** [Makefile:54: check] Error 1
```

I'm using wasmer as my wasm runtime. This error is unrelated to my changes, and hopefully will not be a blocker.

arvid-norlander:
I don't have windows, but it looks like it cmake generate with Ninja instead of make on Windows. I have nothing against ninja, in fact I prefer it. Can we just default to always using ninja for the test across the board (with -GNinja). Or is it not part of the Linux setup?

tests/run.sh
```diff
@@ -54,3 +61,22 @@ for options in -O0 -O2 "-O2 -flto"; do
     done
 done
 cd - >/dev/null
+
+if [[ -n "$wasi_sdk" ]]; then
+    for option in Debug Release; do
+        rm -rf "$testdir/cmake/build/$option"
+        mkdir -p "$testdir/cmake/build/$option"
+        cd "$testdir/cmake/build/$option"
+        cmake \
```
abrown, line 70: Maybe we just need to add a `-G "Unix Makefiles"` around here somewhere to make this work? I'm not strongly opposed to generating Ninja files but since (a) we know we have `make` working in all of the CI agents and (b) this project already used `Makefile`s extensively, maybe we should just keep going with that? Or did I misunderstand something about your comment?

abrown:
> This error is unrelated to my changes, and hopefully will not be a blocker

Yeah, just looks like a different error message. If you use Wasmtime, which the original author probably did, I would expect the error to go away.

arvid-norlander:
This should fix it. Unfortunately I can't start the CI.

tests/run.sh
```diff
@@ -20,12 +20,19 @@ else
     runwasi=""
 fi
 
+if [ $# -gt 1 ]; then
+    wasi_sdk="$2"
+else
+    wasi_sdk=""
+fi
```
sbc100, line 27: Can we just make this a required argument .. perhaps the first one, so the runtime can be the optional second arg?

tests/run.sh
```diff
@@ -54,3 +61,23 @@ for options in -O0 -O2 "-O2 -flto"; do
     done
 done
 cd - >/dev/null
+
+if [[ -n "$wasi_sdk" ]]; then
```
sbc100, line 65: Then you can remove this  `if/fi` block

wasi-sdk.cmake
```diff
@@ -3,6 +3,10 @@
 # This is arbitrary, AFAIK, for now.
 cmake_minimum_required(VERSION 3.4.0)
 
+# Until Platform/WASI.cmake is upstream we need to inject the path to it
+# into CMAKE_MODULE_PATH.
+list(APPEND CMAKE_MODULE_PATH "${WASI_SDK_PREFIX}/share/cmake")
```
sbc100, line 8: Can we use a path relative to the toolchain file itself?  or at least default to a relative path so its not required to always pass the extra WASI_SDK_PREFIX argument?

tests/run.sh
sbc100: Maybe wrap this into a back function such as `test_cmake` ?

arvid-norlander:
Hi, I have been on vacation, but I'll look into your feedback in the coming days.

wasi-sdk.cmake
arvid-norlander: Probably not: When building WASI-SDK itself `WASI_SDK_PREFIX` is not in the correct relative spot compared to this file yet. If I'm interpreting the Makefile logic correctly. But I can give it a try.

And WASI_SDK_PREFIX is already needed further down in the file for finding the compilers as well (so I'm not introducing a new thing). Do we want to make those use relative paths to the cmake file too?

arvid-norlander: Turns out I was wrong, we can do this. I have not changed the use further down for locating the compilers though

arvid-norlander:
Now also rebased onto tip

arvid-norlander:
Not sure what that push from @sbc100  was about (rebase?)

Is this waiting on anything else at this point?

sbc100:
It looks like you still have three commits on top of main somehow (your new one and two existing ones).

sbc100:
> It looks like you still have three commits on top of main somehow (your new one and two existing ones).

Sorry, ignore me.  I forgot that you took #182 as part of this change.

sbc100 (approved):

Issue #336, opened 2023-07-20
Recent macos artifacts are no longer universal binaries


TerrorJack:
```
$ -bash-5.1$ file wasi-sdk-20.9g00cb3b2e5ffc/bin/clang-16
wasi-sdk-20.9g00cb3b2e5ffc/bin/clang-16: Mach-O 64-bit executable x86_64
```

Looks like to me 9a271da9378b407e7234dcbc26ae60fdd80b6d1a is the culprit. There may be a better way to make sure wasi-sdk builds for aarch64 linux while preserving arm64/x86_64 universal binaries for macos as well, I'll look into it a bit.

sbc100:
It looks like 9a271da should be a no-op on Darwin.   Why/how do you think it made any difference?

TerrorJack:
If you compare sizes of `dist-macos-latest` from https://github.com/WebAssembly/wasi-sdk/actions/runs/4953600854 to https://github.com/WebAssembly/wasi-sdk/actions/runs/5224553231, you'll see it almost halved since that commit, suggesting it is the culprit. For some reason, the artifacts and build logs aren't available for the time being, though I suspect it's a transient GitHub issue and we can look into the artifacts later.

I don't think it's an no-op. We pass `make LLVM_CMAKE_FLAGS=...` and for some reason the append logic in the makefile didn't happen, so it seems.

sbc100:
I see, so its a bug in 9a271da and the `LLVM_CMAKE_FLAGS +=` is not working as expected.    Does one of you (@TerrorJack or @alexcrichton) have time if investigate?   Most likely some kind of GNU Make edge case.

alexcrichton:
Ah sorry I don't have time to investigate this. If it's causing issues feel free to revert the commit.

abrown:
#337 closes this.

PR #337, opened 2023-07-27
ci: ensure `LLVM_CMAKE_FLAGS` additions are kept

abrown:
In #321, some OSX-specific `Makefile` additions to `LLVM_CMAKE_FLAGS` were skipped unless `make` is run on a Darwin OS. This allowed building wasi-sdk for aarch64. But, as reported in #336, this also broke arm64/x86_64 universal binaries that are built during CI.

The reason for this is that CI's `main.yml` overrides `LLVM_CMAKE_FLAGS` to add caching but `make` will not append to a variable set on the command line. This changes uses the `override` keyword to append to such a variable, as suggested [here].

[here]: https://www.gnu.org/software/make/manual/html_node/Override-Directive.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ BASH=
 endif
 
 ifeq ($(shell uname),Darwin)
-LLVM_CMAKE_FLAGS += -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" \
+override LLVM_CMAKE_FLAGS += -DCMAKE_OSX_ARCHITECTURES="arm64;x86_64" \
 		    -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12
 endif
 


TerrorJack (approved):
til about the `override` keyword in makefiles, thanks for the fix! macos build log shows `-arch arm64 -arch x86_64` in the llvm build flags so i believe this is a valid fix.

abrown:
Credit to @sunfishcode for recommending `override`!

sbc100 (approved):

abrown:
Ok, looking at the artifacts for this PR's [run](https://github.com/WebAssembly/wasi-sdk/actions/runs/5684370027), `dist-macos-latest` now add up to 118MB prior to a previous [run](https://github.com/WebAssembly/wasi-sdk/actions/runs/5663419211)'s 65.1MB. That seems to be a good sign that this has fixed #336; @TerrorJack, let me know if doesn't!

PR #338, opened 2023-07-28
add shared library support

dicej:
This adds support for building WASI shared libraries per https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md.

For the time being, the goal is to allow "pseudo-dynamic" linking using the Component Model per
https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md. This requires all libraries to be available when the component is created, but still allows runtime symbol resolution via `dlopen`/`dlsym` backed by a static lookup table.  This is sufficient to support Python native extensions, for example.  A complete demo using `wit-component` is available at https://github.com/dicej/component-linking-demo.

This requires https://reviews.llvm.org/D153293, which we will need to backport to LLVM 16 until 17 is released, hence the llvm-D153293-backport.patch file.


.gitmodules
```diff
@@ -1,9 +1,9 @@
 [submodule "src/llvm-project"]
-	path = src/llvm-project
-	url = https://github.com/llvm/llvm-project
+        path = src/llvm-project
+        url = https://github.com/llvm/llvm-project
 [submodule "src/wasi-libc"]
-	path = src/wasi-libc
-	url = https://github.com/CraneStation/wasi-libc
+        path = src/wasi-libc
+        url = https://github.com/dicej/wasi-libc
```
dicej, line 6: I'll switch this back once https://github.com/WebAssembly/wasi-libc/pull/429 is merged.

Makefile
```diff
@@ -194,8 +200,8 @@ build/libcxx.BUILT: build/llvm.BUILT build/compiler-rt.BUILT build/wasi-libc.BUI
 	mkdir -p build/libcxx
 	cd build/libcxx && cmake -G Ninja $(LIBCXX_CMAKE_FLAGS:@PTHREAD@=OFF) \
 		-DCMAKE_SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
-		-DCMAKE_C_FLAGS="$(DEBUG_PREFIX_MAP) $(EXTRA_CFLAGS)" \
-		-DCMAKE_CXX_FLAGS="$(DEBUG_PREFIX_MAP) $(EXTRA_CXXFLAGS)" \
+		-DCMAKE_C_FLAGS="$(DEBUG_PREFIX_MAP) $(EXTRA_CFLAGS) -fPIC" \
```
TerrorJack, line 203: If we are going to apply `-fPIC` uniformly for the time being, a more idiomatic way to do this shall be setting `CMAKE_POSITION_INDEPENDENT_CODE` to `On` in the `wasi-sdk.cmake` and `wasi-sdk-pthread.cmake` files. The flags in `wasi-sdk.cmake` will be used by `compiler-rt` and `libcxx`/`libcxxabi` builds anyway.

dicej: Done, thanks!

Makefile
```diff
@@ -60,6 +60,9 @@ clean:
 
 build/llvm.BUILT:
 	mkdir -p build/llvm
+	# Apply https://reviews.llvm.org/D153293 as a patch until we're ready to
+	# upgrade LLVM (presumably once v17 has been released)
+	(cd $(LLVM_PROJ_DIR) && patch -p1 -N < ../../llvm-D153293-backport.patch) || true
```
dicej, line 65: The annoying thing about this is it leaves the llvm-project submodule in a "dirty" state. I'm not sure if there's a way to reliably avoid that besides copying the whole source tree to a temporary directory and patching that instead, which feels like overkill.

dicej: Of course it won't be necessary at all once we upgrade to LLVM 17.

abrown: You should be able rebase to get LLVM 17.0.1; I guess the patch file can go away?

yamt:
i prefer not to release a binary with __wasm_call_ctors.
cf. https://github.com/dicej/component-linking-demo/issues/3

dicej:
I've rebased onto main and removed the patch.  @yamt does that address your concern regarding `__wasm_call_ctors`?

src/llvm-project
abrown, line 1: @dicej, I think we want to keep this submodule at e19b7d so it stays in sync with [v17.0.1](https://github.com/llvm/llvm-project/releases/tag/llvmorg-17.0.1), right?

dicej: Ugh, yes of course.  I was trying to revert some spurious whitespace changes, and must have messed up the commit.  I'll fix it when back at my computer.  Thanks for pointing it out.

yamt:
> I've rebased onto main and removed the patch. @yamt does that address your concern regarding `__wasm_call_ctors`?

no.
the simplest fix would be to include https://reviews.llvm.org/D156205, which is unfortunately not included in LLVM 17.

wasi-sdk.cmake
```diff
@@ -38,3 +38,5 @@ set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
 set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
+
+set(CMAKE_POSITION_INDEPENDENT_CODE On)
```
yamt, line 42: i'm not sure how these cmake files are supposed to be used to build shared library. can you explain a bit?
is it intended to always build pic even if shared lib is not used? why?

dicej: I think I added `CMAKE_POSITION_INDEPENDENT_CODE` to `LIBCXX_CMAKE_FLAGS` in the Makefile, but it wasn't working (i.e. libc++.so and libc++abi.so still contained position-dependent code), so I added the lines to wasi-sdk.cmake and wasi-sdk-pthread.cmake.  However, I just tried moving it back to Makefile, it that works fine now, so I'll push update.

dicej:
> > I've rebased onto main and removed the patch. @yamt does that address your concern regarding `__wasm_call_ctors`?
> 
> no. the simplest fix would be to include https://reviews.llvm.org/D156205, which is unfortunately not included in LLVM 17.

How does this plan sound to you?

1. Create a new wasi-libc PR to remove the `-nostdlib` parameter from libc.so link command (i.e. let wasm-ld link in crt-reactor.o).
2. Add a llvm-D156205-backport.patch to this PR and apply it as part of the build/llvm.BUILT target.

sbc100:
> > > I've rebased onto main and removed the patch. @yamt does that address your concern regarding `__wasm_call_ctors`?
> > 
> > 
> > no. the simplest fix would be to include https://reviews.llvm.org/D156205, which is unfortunately not included in LLVM 17.
> 

Its possible you could get that patch backported to the 17 branch.. it seems very low risk.

> How does this plan sound to you?
> 
> 1. Create a new wasi-libc PR to remove the `-nostdlib` parameter from libc.so link command (i.e. let wasm-ld link in crt-reactor.o).
> 2. Add a llvm-D156205-backport.patch to this PR and apply it as part of the build/llvm.BUILT target.

dicej:
More generally, I wonder if we should follow the Rust project's example and create a WebAssembly/llvm-project fork as a home for backports (cf. https://github.com/rust-lang/llvm-project).  I do agree that backporting upstream is ideal whenever possible, but it would be nice to have a fallback option that doesn't involve applying patches at build time.

sbc100:
> More generally, I wonder if we should follow the Rust project's example and create a WebAssembly/llvm-project fork as a home for backports (cf. https://github.com/rust-lang/llvm-project). I do agree that backporting upstream is ideal whenever possible, but it would be nice to have a fallback option that doesn't involve applying patches at build time.

Personally, I think the fact that we build wasi-sdk directly from upstream llvm is a really nice feature.  It means the sdk is truly just a vanilla clang.  Nothing more and nothing less.

wasi-libc also supports a range of stable llvm releases which is really nice.  Folks can take wasi-libc and use it with their own pre-built llvm binaries.

If wasi-sdk started to diverge from vanilla llvm I think it could cause confusion and mean that such users would not have the same experience as wasi-sdk users.

yamt:
> Its possible you could get that patch backported to the 17 branch.. it seems very low risk.

i tried to make a backport request, following https://www.llvm.org/docs/GitHub.html#backporting-fixes-to-the-release-branches.
but i couldn't create an issue associated to the milestone.
does it require some special permission?

dicej:
FYI, I've opened https://github.com/WebAssembly/wasi-libc/pull/440 to remove the `-nostdlib` flag from the libc.so link command.  Once that's merged, and the LLVM patch is backported, I can update both submodules in this PR.

tschneidereit:
@sbc100 do you have any advice on the issue @yamt ran into in trying to backport the patch to the 17 branch?

sbc100:
> @sbc100 do you have any advice on the issue @yamt ran into in trying to backport the patch to the 17 branch?

I don't think I've tried to backport anything since the switch to github, but it seems unlikely that they wouldn't accept request from external folks.

@yamt can you link to the issue you created and I can see if I can mark it correctly?

yamt:
> > @sbc100 do you have any advice on the issue @yamt ran into in trying to backport the patch to the 17 branch?
> 
> I don't think I've tried to backport anything since the switch to github, but it seems unlikely that they wouldn't accept request from external folks.
> 
> @yamt can you link to the issue you created and I can see if I can mark it correctly?

https://github.com/llvm/llvm-project/issues/68385
https://github.com/llvm/llvm-project/issues/68386
https://github.com/llvm/llvm-project/issues/68387
(multiple issues just because i made a few attempts)

sbc100:
Looks like I was able to add the Milestone to https://github.com/llvm/llvm-project/issues/68387

yamt:
> Looks like I was able to add the Milestone to [llvm/llvm-project#68387](https://github.com/llvm/llvm-project/issues/68387)

thank you

dicej:
@sbc100 In case you missed it, looks like this is waiting on confirmation from you: https://github.com/llvm/llvm-project-release-prs/pull/773#issuecomment-1807590238

dicej:
I've updated both the `wasi-libc` and `llvm-project` submodules (including llvm-D156205), so I think this is ready to merge.

dicej:
Looks like the CI failure was due to a spurious network issue.  Would someone who has the appropriate permissions please re-run the failed job?

sbc100:
Can we do the llvm update as a separate PR maybe?

yamt:
> Can we do the llvm update as a separate PR maybe?

@dicej do you want to make a separate PR?
or, i can reopen and update https://github.com/WebAssembly/wasi-sdk/pull/359

Makefile
```diff
@@ -107,21 +107,24 @@ build/llvm.BUILT:
...
-		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot
+		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
+		BUILTINS_LIB=$(BUILD_PREFIX)/lib/clang/$(CLANG_VERSION)/lib/wasi/libclang_rt.builtins-wasm32.a \
+		default libc_so
 	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
 		CC=$(BUILD_PREFIX)/bin/clang \
 		AR=$(BUILD_PREFIX)/bin/llvm-ar \
 		NM=$(BUILD_PREFIX)/bin/llvm-nm \
 		SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
+		BUILTINS_LIB=$(BUILD_PREFIX)/lib/clang/$(CLANG_VERSION)/lib/wasi/libclang_rt.builtins-wasm32.a \
```
sbc100, line 123: Why is `BUILTINS_LIB` now needed where it wasn't before?

.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = https://github.com/llvm/llvm-project
 [submodule "src/wasi-libc"]
 	path = src/wasi-libc
-	url = https://github.com/CraneStation/wasi-libc
+	url = https://github.com/WebAssembly/wasi-libc
```
sbc100, line 6: Seems unrelated.. but a good change.

sbc100 (approved):
lgtm now!

(If I had my druthers to wasm-libc updated would also land separately..)

sbc100 (approved):

PR #339, opened 2023-08-09
Add wasi-sdk-pthread.cmake to docker image to allow enabling it with env

anuraaga:
TBH, I'm not sure what to think of this change as it does seem to bring the number of cmake files in this repo to an incredible level. But the docker image is the easiest way to just try building a repo with WASI (e.g., https://github.com/abseil/abseil-cpp/pull/1509), and being able to just `export CMAKE_TOOKCHAIN_FILE=/usr/share/cmake/wasi-sdk-pthread.cmake` would be a great boon.


docker/wasi-sdk.cmake
```diff
@@ -18,6 +18,7 @@ set(CMAKE_AR /usr/bin/llvm-ar-$ENV{LLVM_VERSION})
 set(CMAKE_RANLIB /usr/bin/llvm-ranlib-$ENV{LLVM_VERSION})
 set(CMAKE_C_COMPILER_TARGET ${triple})
 set(CMAKE_CXX_COMPILER_TARGET ${triple})
+set(CMAKE_ASM_COMPILER_TARGET ${triple})
```
anuraaga, line 21: When cribbing from top-level cmake file, I noticed this line was there that hadn't been added here

PR #345, opened 2023-09-01
Do not use GNU find features on FreeBSD

kiancross:
FreeBSD (like MacOS) does not support the `-executable` argument for `find`. This commit checks if the OS is FreeBSD, and if so, uses the same workaround as already in place on MacOS.

A more portable solution might be to test if the `-executable` flag is available, and fallback to `-perm` if not.


sunfishcode:
Thanks!

PR #355, opened 2023-09-27
Set -fno-exceptions in CXXFLAGS

anuraaga:
A while back, the idea to add no-exceptions to CXXFLAGS in the docker image came up but we held up to keep consistency with the cmake files

https://github.com/WebAssembly/wasi-sdk/pull/271#discussion_r1055011577

I thought it would be good to bring it up again with the cmake files also updated. AFAICT, when using wasi-sdk with C++ it almost always ends up the user's responsibility to set this as exceptions aren't supported, so it can be a bit easier by defaulting it

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -48,5 +48,5 @@ ENV AR llvm-ar-${LLVM_VERSION}
 ENV RANLIB llvm-ranlib-${LLVM_VERSION}
 
 ENV CFLAGS --target=wasm32-wasi --sysroot=/wasi-sysroot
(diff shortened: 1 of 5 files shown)


sbc100:
Seems fairly reasonable.  I wonder if this should be the default in clang?

anuraaga:
> I wonder if this should be the default in clang?

Not sure what it would involve but seems like a great idea to me. Would we still consider this PR in the meantime or focus on that?

Also looks like will need https://github.com/WebAssembly/wasi-sdk/pull/354 in for the CI

PR #364, opened 2023-12-11
Build llvm with LLVM_INSTALL_TOOLCHAIN_ONLY

sbc100:
This means that `make install` then only installs things that are meant to be part of the public facing toolchain.

See https://llvm.org/docs/BuildingADistribution.html#difference-between-install-and-install-distribution

The list binaries that get installed now differ as follows:

```
--- old
+++ new
@@ -1,34 +1,72 @@
+addr2line
+amdgpu-arch
+analyze-build
 ar
 c++filt
 clang
 clang++
 clang-17
 clang-apply-replacements
+clang-change-namespace
+clang-check
 clang-cl
 clang-cpp
+clangd
+clang-doc
+clang-extdef-mapping
 clang-format
+clang-include-cleaner
+clang-include-fixer
+clang-linker-wrapper
+clang-move
+clang-offload-bundler
+clang-offload-packager
+clang-pseudo
+clang-query
+clang-refactor
+clang-rename
+clang-reorder-fields
+clang-scan-deps
 clang-tidy
+diagtool
+find-all-symbols
 git-clang-format
+hmaptool
+intercept-build
 ld64.lld
 ld.lld
 lld
 lld-link
 llvm-ar
+llvm-cov
 llvm-cxxfilt
-llvm-dwarfdump
-llvm-mc
+llvm-dwp
+llvm-lib
+llvm-ml
 llvm-nm
 llvm-objcopy
 llvm-objdump
+llvm-pdbutil
+llvm-profdata
 llvm-ranlib
+llvm-rc
+llvm-readobj
 llvm-size
 llvm-strings
 llvm-strip
+llvm-symbolizer
+modularize
 nm
+nvptx-arch
 objcopy
 objdump
+pp-trace
 ranlib
+readelf
 run-clang-tidy
+scan-build
+scan-build-py
+scan-view
 size
 strings
 strip
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,31 +80,12 @@ build/llvm.BUILT:
 		          -DDEFAULT_SYSROOT=../share/wasi-sysroot, \
 			  -DDEFAULT_SYSROOT=$(PREFIX)/share/wasi-sysroot), \
 		     -DDEFAULT_SYSROOT=$(PREFIX)/share/wasi-sysroot) \
-		-DLLVM_INSTALL_BINUTILS_SYMLINKS=TRUE \
+		-DLLVM_INSTALL_TOOLCHAIN_ONLY=ON \
+		-DLLVM_INSTALL_BINUTILS_SYMLINKS=ON \
 		-DLLVM_ENABLE_LIBXML2=OFF \
 		$(LLVM_CMAKE_FLAGS) \
 		$(LLVM_PROJ_DIR)/llvm
-	DESTDIR=$(DESTDIR) ninja $(NINJA_FLAGS) -C build/llvm \
-		install-clang \
-		install-clang-format \
-		install-clang-tidy \
-		install-clang-apply-replacements \
-		install-lld \
-		install-llvm-mc \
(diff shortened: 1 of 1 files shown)


sunfishcode:
The loss of llvm-dwarfdump seems surprising; do you know why isn't that included, when things like llvm-objdump are?

sbc100:
> The loss of llvm-dwarfdump seems surprising; do you know why isn't that included, when things like llvm-objdump are?

The list seems to be hardcoded here: https://github.com/llvm/llvm-project/blob/29ee66f4a0967e43a035f147c960743c7b640f2f/llvm/cmake/modules/AddLLVM.cmake#L1306-L1340

I'm not sure why dwarfdump is not on this list.. perhaps its viewed as an internal debugging tool?   We could make and exception and add it explicitly?  Or ask upstream why its not there already?

sunfishcode:
Would it be difficult to add back in here? Asking upstream may be worth doing, but it'd be nice if we could do something easy for now.

PR #374, opened 2024-01-10
doc: remember to tag wasi-libc during a release

abrown:
This change documents a final step called out in [wasi-libc#461]: tag the wasi-libc repository with the same wasi-sdk version tag.

[wasi-libc#461]: https://github.com/WebAssembly/wasi-libc/issues/461


sunfishcode (approved):
Thanks for writing this up!

PR #377, opened 2024-01-17
Explicitly set LIBCXXABI_USE_LLVM_UNWINDER off

glandium:
Clang trunk has recently turned it on, and also now fails the build when libunwind is not built at the same time.


glandium:
I'm baffled that this breaks things, considering OFF was the default before...

glandium:
Ok, it's broken even without the patch: https://github.com/glandium/wasi-sdk/actions/runs/7550972389/job/20557473630#logs
Something has changed in the environment between the last green run on main and now, presumably wasmtime.

abrown:
Yeah, looks like a WASI signature change. I wonder if we should just pin Wasmtime for now or figure out how to recompile with the new signature.

Changqing-JING:
@abrown Several recent PRs are blocked by the signature change, what's the plan of fixing it?

abrown:
> Ok, it's broken even without the patch: https://github.com/glandium/wasi-sdk/actions/runs/7550972389/job/20557473630#logs Something has changed in the environment between the last green run on main and now, presumably wasmtime.

I believe #382 fixes this so a rebase on top of that (once it merges) _should_ fix the test failure. It's related to the ABI of the preview1-to-preview2 adapter that Wasmtime uses.

glandium:
Merged in #393

PR #380, opened 2024-02-03
Use `apt-get autoremove` to clean up dependencies in Docker

andylizi:
Currently there are left-over packages remain in the docker image:

```
The following packages were automatically installed and are no longer required:
    dirmngr gnupg-l10n gnupg-utils gpg gpg-agent gpg-wks-client gpg-wks-server
    gpgconf gpgsm libassuan0 libksba8 libnpth0 pinentry-curses
Use 'apt autoremove' to remove them.
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,7 +30,7 @@ RUN apt-get update && \
 \
     apt-get update && \
     apt-get install -y clang-${LLVM_VERSION} lld-${LLVM_VERSION} cmake ninja-build make autoconf autogen automake libtool && \
-    apt-get remove -y curl gnupg && \
+    apt-get autoremove -y curl gnupg && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


andylizi:
Strange, the CI is failing. I don't think this is caused by this PR?

```
 Error: failed to run main module `abort.c.---.wasm`
 
 Caused by:
-    0: failed to invoke `run` function
-    1: error while executing at wasm backtrace:
-    2: wasm trap: wasm `unreachable` instruction executed
+    0: import `wasi:cli/exit@0.2.0-rc-2023-12-05` has the wrong type
+    1: instance export `exit` has the wrong type
+    2: expected func found nothing
make: *** [Makefile:54: check] Error 1
```

abrown (approved):
Looks good to me. Once #382 merges, I believe we should be able to rebase this to fix the unrelated build failure.

PR #382, opened 2024-02-06
Fix cmake_build_type typo

Changqing-JING:
The cmake_build_type is wrongly spelled. When cmake receives as an invalid cmake_build_type, it won't add optimization flags such as `-O2` to compiler command. Then the complication is unoptimized.


abrown (approved):
Good catch! I confirmed from the CMake [docs](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) that this is the right change and searched through the Makefile to check that there were no other typos of this kind.

Changqing-JING:
@abrown 
I found the abort.c CI error which blocks recent PRs is not difficult to fix.
The problem is just the wasmtime is updated the 17.0.0 but the `wasi_snapshot_preview1.command.wasm` is still 16.0.0, which leads to the version mismatch. By updating `wasi_snapshot_preview1.command.wasm` to 17.0.0 can solve this problem

Changqing-JING:
@abrown ping ping, could you help to approve the CI. The CI error should be fixed by this PR

abrown:
@Changqing-JING thanks for the reminder to look at this again and good find on the adapter incompatibility! I thought it was something related to that from a recent wasi-libc PR.

PR #386, opened 2024-03-02
Start renaming preview1 to p1 and preview2 to p2

alexcrichton:
This commit is a reflection of WebAssembly/wasi-libc#478 into this repository where a few changes are happening:

* A new `wasm32-wasip1` sysroot is prepared matching `wasm32-wasi`
* A new `wasm32-wasip1-threads` sysroot is prepared matching `wasm32-wasi-threads`
* The `wasm32-wasi-preview2` target is renamed `wasm32-wasip2`

I've done a bit of makefile refactoring to deduplicate things a bit now that there's a number of targets being built.

The long-term goal would be to remove the `wasm32-wasi` and `wasm32-wasip1-threads` targets, but that's not proposed just yet at this time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,26 +107,22 @@ build/llvm.BUILT:
 		llvm-config
 	touch build/llvm.BUILT
 
+# Flags for running `make` in wasi-libc
+# $(1): the target that's being built
+WASI_LIBC_MAKEFLAGS = \
+	-C $(ROOT_DIR)/src/wasi-libc \
+	CC=$(BUILD_PREFIX)/bin/clang \
+	AR=$(BUILD_PREFIX)/bin/llvm-ar \
+	NM=$(BUILD_PREFIX)/bin/llvm-nm \
+	SYSROOT=$(BUILD_PREFIX)/share/wasi-sysroot \
+	TARGET_TRIPLE=$(1)
+
 build/wasi-libc.BUILT: build/compiler-rt.BUILT
-	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
-		CC=$(BUILD_PREFIX)/bin/clang \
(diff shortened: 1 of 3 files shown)


alexcrichton:
I'll note that I'm opening this as a draft here because https://github.com/WebAssembly/wasi-libc/pull/478 should land first. In the meantime I've updated the submodule in this PR to point at my fork of wasi-libc, which should not land in this repository (it's just here for testing in CI and showing what this will look like)

alexcrichton:
With https://github.com/WebAssembly/wasi-libc/pull/478 now landed I've updated this to the official submodule and this is no longer a draft.

Makefile
```diff
@@ -156,11 +152,15 @@ build/compiler-rt.BUILT: build/llvm.BUILT
 	DESTDIR=$(DESTDIR) ninja $(NINJA_FLAGS) -C build/compiler-rt install
 	# Install clang-provided headers.
 	cp -R $(ROOT_DIR)/build/llvm/lib/clang $(BUILD_PREFIX)/lib/
+	cp -R $(BUILD_PREFIX)/lib/clang/17/lib/wasi $(BUILD_PREFIX)/lib/clang/17/lib/wasip1
+	cp -R $(BUILD_PREFIX)/lib/clang/17/lib/wasi $(BUILD_PREFIX)/lib/clang/17/lib/wasip2
```
sunfishcode, line 156: Can these use `$(CLANG_VERSION)` instead of hard-coding `17`?

alexcrichton: Ah yes indeed! I didn't look hard enough to find that.

yamt:
> The long-term goal would be to remove the wasm32-wasi and wasm32-wasip1-threads targets, but that's not proposed just yet at this time.

do you mean wasm32-wasi-threads?

PR #388, opened 2024-03-02
Add target-prefixed executables to the installation `bin` directory

alexcrichton:
Currently the `bin` directory installed by wasi-sdk is not currently suitable for putting in `$PATH` in all cases because it can shadow a system-installed `clang` executable which is intended for native binaries. For Linux distributions with gcc-based cross-compilers I've often seen the pattern where they're installed as `$target-gcc` and so I've taken a leaf out of their books to do that here as well.

This commit adds, currently alongside the preexisting `clang` executable, target-prefixed executables such as `wasm32-wasi-clang` and `wasm32-wasi-clang++`. These executables are copies of a Rust program whose source is added here as well, where the program infers from its own executable name what to pass to `clang` and such. I'll note that a Rust program was chosen here over a shell script to have this technique work on Windows too.

I'll also note that this doesn't fix the problem of "add the wasi-sdk bin dir to PATH" because `clang` and other prominent executables are still there. My hope though is that this opens up a future refactoring for doing so.


alexcrichton:
I should say that another motivation for this is that we'll want the preview2 target to output a component by default, not a core module, and for that purpose rustc will use [`wasm-component-ld`](https://github.com/alexcrichton/wasm-component-ld) and I think we'll want to do the same here. To help configure this by default a `wasm32-wasip2-clang` shim executable seems like a good place to inject a `-fuse-ld=/....` argument to `wasm-component-ld`.

alexcrichton:
Ah ok, I think I'm too used to distros and gcc things then. If it's not idiomatic for clang to have target-prefixes I agree this needs to change.

I suppose then this raises a bit of a larger question of if there's a longer-term installation plan for wasi-sdk perhaps? For example is the goal on macOS, for example, to enable `brew install wasi-sdk`? Or for Linux users to `package-manager install wasi-sdk`? In those situations wasi-sdk is more of a sysroot than a build of clang I suppose? That would further motivate getting any "special logic" into Clang itself (e.g. `wasm-component-ld` as a linker for `wasm32-wasip2`).

Do you know how feasible it would be to modify Clang to support `wasm32-wasip2` and `wasm-component-ld`? I think it'd be best to get that change in the installation of wasi-sdk itself sooner-rather-than-later to avoid ossifying the output of `wasm32-wasip2` as a core wasm module as opposed to a component.

Also, yeah, I'm happy to make a `prefixed-bin` directory with shims pointing to `clang` and `clang++`.

sbc100:
> I'll also note that this doesn't fix the problem of "add the wasi-sdk bin dir to PATH" because `clang` and other prominent executables are still there. My hope though is that this opens up a future refactoring for doing so.

I think as long as you add wasi-sdk to the *end* of the `$PATH` this should work fine right?  When `clang` would resolve to the system clang and wasm32-wasi-clang would resolve to wasi-sdk

sbc100:
> It isn't really idiomatic for clang to have target-prefixed executables, because clang has the ability to support multiple targets in the same build

I'm not totally sure that is totally true.   I believe there is precedent out there for target-prefixed clang binaries.  For example android NDK ship no less that 57 target prefixed versions of clang++:

```
$ find android-ndk-r26c/ -iname *clang++ | head
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android21-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android27-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi33-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi32-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android23-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi22-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/x86_64-linux-android23-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android21-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android22-clang++
android-ndk-r26c/toolchains/llvm/prebuilt/linux-x86_64/bin/i686-linux-android26-clang++
$ find android-ndk-r26c/ -iname *clang++ | wc -l 
57
```

I believe the mechansim for setting the default target based on the executable name is built into clang itself, so these can all by symlinks or copies of the primary `clang++` executable.

I've personally found it useful in the past to be able do `CC=myarch-clang` rather than `CC=clang --target=myarch` (IIRC some build systems like CC to be just an executable).

Another advantage of `myarch-clang` is that you can put the directory contains those binaries at the end if your path while still allowing `/usr/bin/clang` to be found first when the arch prefix is not specified.    So you can do things like `CC=myarch-clang HOST_CC=clang`

yamt:
as clang recognizes prefixes as @sbc100 said, i suppose it makes sense for us to ship prefixed symlinks in a separate bin dir.
```
spacetanuki% /opt/wasi-sdk-21.0/bin/clang -print-effective-triple
wasm32-unknown-wasi
spacetanuki% ln -s /opt/wasi-sdk-21.0/bin/clang wasm64-clang
spacetanuki% ./wasm64-wasi-clang -print-effective-triple
wasm64-unknown-wasi
spacetanuki% 
```

agoode:
Prefixed binaries will help simplify https://github.com/MLton/mlton/pull/550. I appreciate this proposed improvement.

sunfishcode:
Thanks @sbc100! I was unaware clang had that, and that other cross-sdks used it that way. That sounds like a good approach to me.

loganek:
> I should say that another motivation for this is that we'll want the preview2 target to output a component by default, not a core module, and for that purpose rustc will use [wasm-component-ld](https://github.com/alexcrichton/wasm-component-ld) and I think we'll want to do the same here.

@alexcrichton  I understand the motivation behind having a component as a default one for the preview2, but will there be an option to use wasm-ld directly for rustc/clang? I think there are usecases where people just want to build a core module with preview2 canonical ABI only, so having a flag to disable component model would be really helpful.

alexcrichton:
> I think as long as you add wasi-sdk to the end of the $PATH this should work fine right?

If executing target-prefixed executables, yes, but IMO that's pretty brittle. In the long term I still think it'd be good to have a directory that can be unambiguously added to `$PATH`. (I'll note I'm not proposing that in this PR though)

>  I looks like we could set CLANG_LINKS_TO_CREATE to a list of aliases that we want.

Thanks for pointing that out! (as well as the precedent and code Clang already has to handle this!)

I've now updated this PR to using this strategy instead to install symlinks for each target.

> will there be an option to use wasm-ld directly for rustc/clang?

Yes both clang and rustc allow configuring the linker, there's no need to use the built-in default. And also, yes, my plan is to add a flag of some sort to `wasm-component-ld` to emit a core module instead of a component. That'll likely be an integral part of "dynamic" linking a la what componentize-py does today.

Makefile
```diff
@@ -51,16 +53,29 @@ default: build
...
+# Add target-prefixed versions of `clang` and `clang++` so they can be used
+# without `--target` as it's auto-inferred from the executable name by clang.
+CLANG_LINKS_TO_CREATE += $(foreach target,$(TARGETS),$(target)-clang)
+CLANG_LINKS_TO_CREATE += $(foreach target,$(TARGETS),$(target)-clang++)
+
+# Small helper to create a `join-with` function that can join elements of a
+# list with a defined separator.
+noop =
+space = $(noop) $(noop)
+join-with = $(subst $(space),$1,$(strip $2))
```
sbc100, line 73: Shame that we have to resort for this kind of Makefile magic, but I am not well versed enough to suggest anything better :)

sbc100 (commented):
Awesome!  Nice to see that CLANG_LINKS_TO_CREATE works.

sbc100 (approved):
Can you update the PR description now that a new technique is being used?

Makefile
alexcrichton: Agreed :(

I went through a number of iterations of "copy some bits and pieces from stack overflow" until I found this which ended up working well enough.

alexcrichton:
Ah yes indeed, and thanks for the review!

alexcrichton:
For posterity here: For the wasm32-wasip2 targets and using `wasm-component-ld` by default I submitted https://github.com/llvm/llvm-project/pull/84569 which was backported to 18.1.x in https://github.com/llvm/llvm-project/pull/85802 which I believe will target the April 2 release of 18.1.2 for LLVM.

Issue #372, opened 2024-01-08
How to tell what version of the WASI-SDK is installed?


brettcannon:
E.g., is there a text file somewhere I can have users look at to tell me what they tried using? (Context: https://github.com/python/cpython/issues/113516).

abrown:
Reading through the issue, I see what you mean. Usually it's apparent if the user shows the full path to Clang, e.g., `/opt/wasi-sdk-20.0/bin/clang`. But it seems reasonable to add a version file somewhere for this kind of situation.

sbc100:
Some kind of version.h header might be good idea

PR #392, opened 2024-03-07
Replace the version scripts

abrown:
In the past, I've had trouble keeping the current set of version scripts
to correctly output the version, especially in light of how we append
Git suffixes for a non-tagged commit and dirty state. This change
replaces those scripts with a single Python one which, though much more
wordy than the previous one, may be easier for contributors to read and
modify. The original scripts relied on Perl; this one relies on Python,
which seems like a fair exchange. Having a single script also makes it
easier to solve https://github.com/WebAssembly/wasi-sdk/issues/372.

As noted in https://github.com/WebAssembly/wasi-sdk/issues/372, some users may need to identify which version of
wasi-sdk they are using after it has been downloaded. There are many
ways to solve this, but the one I chose here is to dump the wasi-sdk
version into a `VERSION` file and follow that up with several dependency
versions that may be helpful to note when troubleshooting. Ideally
someone could paste the contents of that file when filing a bug. If we
adopt this approach, this fixes https://github.com/WebAssembly/wasi-sdk/issues/372.

### DIFF
--- a/docker_build.sh
+++ b/docker_build.sh
@@ -1,7 +1,18 @@
 #!/bin/sh
 set -ex
+
 echo "Building the docker image"
-docker build -t wasi-sdk-builder:latest .
+docker build \
+    --build-arg UID=$(id -u) --build-arg GID=$(id -g) \
+    -t wasi-sdk-builder:latest .
+
 echo "Building the package in docker image"
 mkdir -p ~/.ccache
-docker run --rm -v "$PWD":/workspace -v ~/.ccache:/root/.ccache -e NINJA_FLAGS=-v --workdir /workspace --tmpfs /tmp:exec wasi-sdk-builder:latest make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON
+docker run --rm \
+    --user $(id -u):$(id -g) \
+    -v "$PWD":/workspace:Z \
+    -v ~/.ccache:/home/builder/.ccache:Z \
+    -e NINJA_FLAGS=-v \
+    --tmpfs /tmp:exec \
+    wasi-sdk-builder:latest \
+    make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON
--- a/version.py
+++ b/version.py
@@ -5,9 +5,10 @@
 #
 # Usage: version [wasi-sdk|llvm|llvm-major|dump] [--llvm-dir=<non-project dir>]
 
-import sys
 import argparse
+import os
 import subprocess
+import sys
 
 # The number of characters to use for the abbreviated Git revision.
 GIT_REF_LEN = 12
@@ -75,7 +76,11 @@ def parse_cmake_set(line):
 
 
 def llvm_cmake_version(llvm_dir):
-    with open(f'{llvm_dir}/llvm/CMakeLists.txt') as file:
+    path = f'{llvm_dir}/cmake/Modules/LLVMVersion.cmake'
+    if not os.path.exists(path):
+        # Handle older LLVM versions; see #399.
+        path = f'{llvm_dir}/llvm/CMakeLists.txt'
+    with open(path) as file:
         for line in file:
             line = line.strip()
             if line.startswith('set(LLVM_VERSION_MAJOR'):
--- a/version.py
+++ b/version.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python3
+
+# This script finds and prints the various versions in this project: wasi-sdk
+# itself, LLVM, and the Git revisions of dependencies.
+#
+# Usage: version [wasi-sdk|llvm|llvm-major|dump] [--llvm-dir=<non-project dir>]
+
+import sys
+import argparse
+import subprocess
+
+# The number of characters to use for the abbreviated Git revision.
+GIT_REF_LEN = 12
+
+
+def exec(command, cwd=None):
+    result = subprocess.run(command, stdout=subprocess.PIPE,
+                            universal_newlines=True, check=True, cwd=cwd)
+    return result.stdout.strip()
+
+
+def git_commit(dir='.'):
+    return exec(['git', 'rev-parse', f'--short={GIT_REF_LEN}', 'HEAD'], dir)
+
+
+def parse_git_version(version):
+    # Parse, e.g.: wasi-sdk-21-0-g317548590b40+m
+    parts = version.replace('+', '-').split('-')
+    assert parts.pop(0) == 'wasi'
+    assert parts.pop(0) == 'sdk'
+
+    major, minor = parts.pop(0), parts.pop(0)
+    git = None
+    dirty = False
+
+    if parts:
+        # Check: git|dirty.
+        next = parts.pop(0)
+        if next == 'm':
+            dirty = True
+        else:
+            git = next[1:]
+
+        # Check: dirty.
+        if parts:
+            assert parts.pop(0) == 'm', f'expected dirty flag: +m'
+            dirty = True
+
+    assert not parts, f'unexpected suffixes: {parts}'
+    return major, minor, git, dirty
+
+
+# Some inline tests to check Git version parsing:
+assert parse_git_version(
+    'wasi-sdk-21-0-g317548590b40+m') == ('21', '0', '317548590b40', True)
+assert parse_git_version('wasi-sdk-21-2+m') == ('21', '2', None, True)
+assert parse_git_version(
+    'wasi-sdk-23-0-g317548590b40') == ('23', '0', '317548590b40', False)
+
+
+def git_version():
+    version = exec(['git', 'describe', '--long', '--candidates=999',
+                    '--match=wasi-sdk-*', '--dirty=+m', f'--abbrev={GIT_REF_LEN}'])
+    major, minor, git, dirty = parse_git_version(version)
+    version = f'{major}.{minor}'
+    if git:
+        version += f'g{git}'
+    if dirty:
+        version += '+m'
+    return version
+
+
+def parse_cmake_set(line):
+    return line.split(' ')[1].split(')')[0]
+
+
+def llvm_cmake_version(llvm_dir):
+    with open(f'{llvm_dir}/llvm/CMakeLists.txt') as file:
+        for line in file:
+            line = line.strip()
+            if line.startswith('set(LLVM_VERSION_MAJOR'):
+                llvm_version_major = parse_cmake_set(line)
+            elif line.startswith('set(LLVM_VERSION_MINOR'):
+                llvm_version_minor = parse_cmake_set(line)
+            elif line.startswith('set(LLVM_VERSION_PATCH'):
+                llvm_version_patch = parse_cmake_set(line)
+    return llvm_version_major, llvm_version_minor, llvm_version_patch
+
+
+def main(action, llvm_dir):
+    if action == 'wasi-sdk':
+        print(git_version())
+    elif action == 'llvm':
+        major, minor, path = llvm_cmake_version(llvm_dir)
+        print(f'{major}.{minor}.{path}')
+    elif action == 'llvm-major':
+        major, _, _ = llvm_cmake_version(llvm_dir)
+        print(major)
+    elif action == 'dump':
+        print(git_version())
+        print(f'wasi-libc: {git_commit("src/wasi-libc")}')
+        print(f'llvm: {git_commit(llvm_dir)}')
+        major, minor, path = llvm_cmake_version(llvm_dir)
+        print(f'llvm-version: {major}.{minor}.{path}')
+        print(f'config: {git_commit("src/config")}')
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
+        description='Print the various kinds of versions in wasi-sdk')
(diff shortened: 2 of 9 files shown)


abrown:
For reference, `./version dump` emits something like the following, which gets placed in a `VERSION` file in the wasi-sdk installation directory:

```
21.11gdb977fd767ca
wasi-libc: c9c7d0616e0f
llvm: 6009708b4367
llvm-version: 17.0.6
config: f992bcc08219
```

version
```diff
@@ -0,0 +1,119 @@
+#!/usr/bin/env python
```
sbc100, line 1: Can we call this `version.py`, even though its executable via the `#!`?

version
```diff
@@ -0,0 +1,119 @@
...
+                llvm_version_major = parse_cmake_set(line)
+            elif line.startswith('set(LLVM_VERSION_MINOR'):
+                llvm_version_minor = parse_cmake_set(line)
+            elif line.startswith('set(LLVM_VERSION_PATCH'):
+                llvm_version_patch = parse_cmake_set(line)
+    return llvm_version_major, llvm_version_minor, llvm_version_patch
+
+
+def parse_cmake_set(line):
+    return line.split(' ')[1].split(')')[0]
```
sbc100, line 98: The order of the functions in this file seems be kind of backwards.   I would normally expect to see functions declared before their first use.   e.g. `parse_cmake_set` before `llvm_cmake_version` and `main` at the very end.

Not a big deal I guess.

version
```diff
@@ -0,0 +1,119 @@
...
+
+
+def exec(command, cwd='.'):
```
sbc100, line 101: `cwd=None` matches the `subprocess.run` flag I think.

version
```diff
@@ -0,0 +1,119 @@
...
+                        choices=['wasi-sdk', 'llvm', 'llvm-major', 'dump'],
+                        nargs='?',
+                        default='wasi-sdk',
+                        help='Which kind of version to print (default: wasi-sdk).')
+    parser.add_argument('--llvm-dir',
+                        nargs='?',
+                        default='src/llvm-project',
+                        help='Override the location of the LLVM source directory (default: src/llvm-project).')
+    args = parser.parse_args()
+    main(args.action, args.llvm_dir)
```
sbc100, line 119: `sys.exit(0)` here.

version
```diff
@@ -0,0 +1,119 @@
...
+assert parse_git_version(
+    'wasi-sdk-23-0-g317548590b40') == ('23', '0', '317548590b40', False)
+
+
+def git_commit(dir='.'):
+    return exec(['git', 'rev-parse', f'--short={GIT_REF_LEN}', 'HEAD'], dir)
+
+
+def llvm_cmake_version(llvm_dir):
+    with open(f'{llvm_dir}/llvm/CMakeLists.txt', 'r') as file:
```
sbc100, line 85: 'r' is the default so you can drop it

version
```diff
@@ -0,0 +1,119 @@
...
+        major, minor, path = llvm_cmake_version(llvm_dir)
+        print(f'llvm-version: {major}.{minor}.{path}')
+        print(f'config: {git_commit("src/config")}')
+
+
+def git_version():
+    version = exec(['git', 'describe', '--long', '--candidates=999',
+                    '--match=wasi-sdk-*', '--dirty=+m', f'--abbrev={GIT_REF_LEN}'])
+    major, minor, git, dirty = parse_git_version(version)
+    version = f'{major}.{minor}'
```
sbc100, line 37: f-strings are a python3 feature so the #! above should say `python3`

version
```diff
@@ -0,0 +1,119 @@
...
+    elif action == 'llvm-major':
+        major, _, _ = llvm_cmake_version(llvm_dir)
+        print(major)
+    elif action == 'dump':
+        print(git_version())
+        print(f'wasi-libc: {git_commit("src/wasi-libc")}')
+        print(f'llvm: {git_commit(llvm_dir)}')
+        major, minor, path = llvm_cmake_version(llvm_dir)
+        print(f'llvm-version: {major}.{minor}.{path}')
+        print(f'config: {git_commit("src/config")}')
```
sbc100, line 30: Do we need this `dump` feature?  I assume it just for debugging?

abrown: It's for solving #372: if you look at the `Makefile` you can see I dump this content into a `VERSION` file that would be included in future distributions of wasi-sdk.

version
abrown: Sure, I reordered in C-like order.

version
abrown: I did this... but now I'm curious why you think this is necessary?

version
abrown: Yup, tried to make this script `python2`-compatible but I forgot about the f-strings! Fixed in 7a4f8f6.

version
sbc100: Not necessary.  I just always prefer main functions to exit explicitly rather than fall off the end of the script.

In fact the pattern i use almost everywhere is:

```
if __name__ == '__main__':
  sys.exit(main(sys.argv[1:]))
```

Then have main return 0 at the end.    That way you can write tests for main itself, and it keeps the top level code down to just a single line.

But thats maybe just my personal preference.

abrown: Nice; makes sense.

abrown:
Looks like the Docker issue, when replicated locally, is due to the `bionic` (18.04) image having only Python 3.6, which does not include the `capture_output` parameter I used in `version.py`. `capture_output` was added in Python 3.7.

version.py
```diff
@@ -0,0 +1,121 @@
...
+import sys
+import argparse
+import subprocess
+
+# The number of characters to use for the abbreviated Git revision.
+GIT_REF_LEN = 12
+
+
+def exec(command, cwd=None):
+    result = subprocess.run(command, capture_output=True, text=True, cwd=cwd)
```
sbc100, line 17: You can use `stdout=PIPE` instead I think

sbc100: In fact `stdout=PIPE` is better in this case I think since `capture_output` will redirect (and potentially hide) stderr message which could be useful for debugging.

abrown: Good suggestion, though I ended up just adding some error-handling code here anyways that captures `stderr`.

version.py
```diff
@@ -0,0 +1,126 @@
...
+import subprocess
+
+# The number of characters to use for the abbreviated Git revision.
+GIT_REF_LEN = 12
+
+
+def exec(command, cwd=None):
+    result = subprocess.run(
+        command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, cwd=cwd)
+    if result.returncode != 0:
```
sbc100, line 19: might be simpler to just add `check=True`, but lgtm either way.

abrown: And drop the `stderr=subprocess.PIPE`, right?

sbc100: Right that would be my recommendation for keeping this code simpler and letting python itself report the errors.

docker_build.sh
```diff
@@ -1,7 +1,14 @@
 #!/bin/sh
 set -ex
+
 echo "Building the docker image"
 docker build -t wasi-sdk-builder:latest .
+
+if ! git config safe.directory | grep --quiet /workspace; then
+    echo "Setting up Git safe directory"
+    git config --global --add safe.directory /workspace
+fi
```
sbc100, line 10: why is this needed?

abrown: The original had it:

https://github.com/WebAssembly/wasi-sdk/blob/317548590b402d622194416094fe0e392b725fe2/version.sh#L3

I didn't know why, but now I do: basically when running any `git` command inside a Docker container the user is different than usual so Git thinks something fishy is going on. Git will print something like:

```
fatal: detected dubious ownership in repository at '/workspace'
To add an exception for this directory, call:

        git config --global --add safe.directory /workspace
```

So the original author of that script inserted the line Git suggests into the `version.sh` script. If you have built wasi-sdk any number of times, you may want to check your `.gitconfig` as it could be replete with these exemptions.

sbc100: So an alternative fix here would me to make sure the `/workspace` (inside the container) appeared to be own by the current user?

Is there maybe a simple way to do that?  Modifying the `.gitconfig` that lives outside of the container seem like a bit of a hack.

abrown: Here's an excerpt from the logs. This build job continues to fail:

```
2024-03-08T23:15:33.7528610Z #8 writing image sha256:54b42a06e7203c0f48481b8ed35206e73b76632b7c8fa05f8fff21636bca69de done
2024-03-08T23:15:33.7531519Z #8 naming to docker.io/library/wasi-sdk-builder:latest done
2024-03-08T23:15:33.7532287Z #8 DONE 2.4s
2024-03-08T23:15:33.7574889Z + git config safe.directory
2024-03-08T23:15:33.7576131Z + grep --quiet /workspace
2024-03-08T23:15:33.7591487Z + echo Setting up Git safe directory
2024-03-08T23:15:33.7592129Z Setting up Git safe directory
2024-03-08T23:15:33.7592946Z + git config --global --add safe.directory /workspace
2024-03-08T23:15:33.7605229Z + echo Building the package in docker image
2024-03-08T23:15:33.7605980Z + mkdir -p /home/runner/.ccache
2024-03-08T23:15:33.7606384Z Building the package in docker image
2024-03-08T23:15:33.7617564Z + docker run --rm -v /home/runner/work/wasi-sdk/wasi-sdk:/workspace -v /home/runner/.ccache:/root/.ccache -e NINJA_FLAGS=-v --workdir /workspace --tmpfs /tmp:exec wasi-sdk-builder:latest make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON
2024-03-08T23:15:34.0117006Z fatal: detected dubious ownership in repository at '/workspace'
2024-03-08T23:15:34.0118288Z To add an exception for this directory, call:
2024-03-08T23:15:34.0118757Z 
2024-03-08T23:15:34.0119150Z 	git config --global --add safe.directory /workspace
2024-03-08T23:15:34.0124011Z Traceback (most recent call last):
2024-03-08T23:15:34.0124634Z   File "./version.py", line 121, in <module>
2024-03-08T23:15:34.0125270Z     main(args.action, args.llvm_dir)
2024-03-08T23:15:34.0125825Z   File "./version.py", line 92, in main
2024-03-08T23:15:34.0126368Z     print(git_version())
2024-03-08T23:15:34.0126865Z   File "./version.py", line 63, in git_version
2024-03-08T23:15:34.0127760Z     '--match=wasi-sdk-*', '--dirty=+m', f'--abbrev={GIT_REF_LEN}'])
2024-03-08T23:15:34.0128517Z   File "./version.py", line 18, in exec
2024-03-08T23:15:34.0128924Z     universal_newlines=True, check=True, cwd=cwd)
2024-03-08T23:15:34.0129390Z   File "/usr/lib/python3.6/subprocess.py", line 438, in run
2024-03-08T23:15:34.0129824Z     output=stdout, stderr=stderr)
2024-03-08T23:15:34.0130809Z subprocess.CalledProcessError: Command '['git', 'describe', '--long', '--candidates=999', '--match=wasi-sdk-*', '--dirty=+m', '--abbrev=12']' returned non-zero exit status 128.
2024-03-08T23:15:34.0746203Z mkdir -p build/llvm
```

Maybe for some reason the command must run inside the Docker image?!

sbc100: the problem there is the `--global` in the git config command.  that will write to your `$HOME/.gitconfig` which is not inside the container image.   Removing the `--global` might help?   (But we have more than one repository here so it might need to be run on each repo)

sbc100: Maybe better to look into UID mapping to avoid the git warning.

abrown: Ok, I looked at this again and tried what I think you meant by user mapping in [91c48f0](https://github.com/WebAssembly/wasi-sdk/pull/392/commits/91c48f0982a31d5988b73f3a822c44d27cbc3a6f). Basically what that means is that `docker_build.sh` will now run something like: `docker run --rm --user 1000:1000 -v /home/abrown/Code/wasi-sdk:/workspace:Z -v /home/abrown/.ccache:/root/.ccache:Z -e NINJA_FLAGS=-v --workdir /workspace --tmpfs /tmp:exec wasi-sdk-builder:latest make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON` (note the additional `:Z` and the `--user` flag). But now I get a different error related to the ccache ([here](https://github.com/WebAssembly/wasi-sdk/actions/runs/8302548419/job/22725057734?pr=392#step:10:1107):

```
ccache: error: Failed to create directory //.ccache/tmp: Permission denied
```

abrown: Well, it looks like ce47f32 is enough to do the user mapping @sbc100 recommended. This essentially supercedes #395, allowing me at least to run the `docker_build.sh` script locally. I am slightly concerned that the Docker images stored by GitHub which have a few downloads each will now have an imprinted UID from GitHub CI; does anyone care about that?

abrown:
@sbc100, @sunfishcode: I rebased this change and re-split the commits as atomic changes on top of `main`. To solve the Docker build issues, this PR now incorporates the essence of #395 (closing now...). This means that now, unlike what we currently have in `main`, I am able to run the `docker_build.sh` script locally.

sbc100 (approved):

PR #394, opened 2024-03-08
Update base for Docker build image

abrown:
I came across this in #392, where it became clear that Ubuntu's 18.04 release (now several years old) was using quite ancient versions of certain tools.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # Use a relatively old/stable distro here to maximize the supported platforms
-# and avoid depending on more recent version of, say, libc.
-# Here we choose Bionic 18.04.
-FROM ubuntu:bionic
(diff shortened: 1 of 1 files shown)


sbc100:
I think the idea is that we want to continue to support Ubuntu's 18.04 (bionic) users.   At least we should probably not stop supporting them unless we have good reason to.

abrown:
Ah, I don't think I understood your `Dockerfile` comment correctly... but I get it now &mdash; maintain backwards compatibility for as long as possible.

alexcrichton:
One thing I'll note is that in rust-lang/rust we ran into toolchain issues a lot and the solution was that while we'd start with an ancient docker container we'd download and build newer versions of tools needed, e.g. a newer git, cmake, openssl, clang, etc.

PR #395, opened 2024-03-08
Allow Docker to run locally in more places

abrown:
These are the changes I had to make to get the `docker_build.sh` script to run locally on my Fedora 39 machine with Docker 24.0.5. I believe the main issue here is one of permissions, which may not be present in CI due to them being elevated (?).


abrown:
Superseded by #392.

PR #400, opened 2024-03-16
Update LLVM 18.1.1

PiCoPress:
Update LLVM v18.1.1, and perhaps fixed submodule (src/config).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,18 +3,23 @@
 # Here we choose Bionic 18.04.
 FROM ubuntu:bionic
 
+# We want to use the same UID/GID of the external user to avoid permission
+# issues. See the user setup at the end of the file.
+ARG UID=1000
+ARG GID=1000
+
 RUN apt-get update \
- && apt-get install -y --no-install-recommends \
-        ccache \
-        curl \
-        ca-certificates \
-        build-essential \
-        clang \
-        python3 \
-        git \
-        ninja-build \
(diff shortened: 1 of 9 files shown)


sbc100:
So here we are updating from 18.1.0rc to 18.1.1?   (Can you mention both these versions in the PR title).  

I'm not sure why ever pinned to an `rc` version but its good that we are now back to an official release.

alexcrichton:
I don't believe an rc was used, it was a [mistake in the LLVM release process](https://github.com/llvm/llvm-project/issues/84168).

.gitmodules
```diff
@@ -7,3 +7,5 @@
 [submodule "src/config"]
 	path = src/config
 	url = https://git.savannah.gnu.org/git/config.git
+[submodule "src/config/"]
```
PiCoPress, line 10: I found dup module definition at .gitmodules, so it should be removed.
For me, previous src/config repository has an issue which git can't fetch from, so I altered its mirror what yields expected result.
```...url/cgit/... -> ...url/git/....```

PiCoPress:
Is there a benefit to accept LLVM 18.1.2?

alexcrichton:
As a heads up I've opened https://github.com/WebAssembly/wasi-sdk/pull/402 to update to 18.1.2 which additionally addresses changes necessary to keep things working with components.

PR #402, opened 2024-03-29
Update LLVM to 18.1.2 and add wasm-component-ld

alexcrichton:
This commit has two intertwined changes within it. First the LLVM submodule is updated to the 18.1.2 release branch. This alone would cause CI and tests to fail due to differing behavior for the `wasm32-wasip2` target. To fix these test failures the `wasm-component-ld` tool is added to the build. This tool is a Rust-written tool and installed via `cargo install` as part of the build at a pinned version written in the `Makefile`. This linker, used for components, is then used for the `wasm32-wasip2` target.

Tests and CI are then updated to skip the need to have `wasm-tools` or the adapter for WASI when making components since that's now the job of `wasm-component-ld`. This then necessitated some changes to tests too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,9 @@ build/llvm.BUILT:
 		llvm-config
 	touch build/llvm.BUILT
 
+# Build the `wasm-component-ld` linker from source via `cargo install`. This is
+# used for the `wasm32-wasip2` target natively by Clang. Note that `--root`
+# passed to `cargo install` will place it in the output directory automatically.
 build/wasm-component-ld.BUILT: build/llvm.BUILT
 	cargo install wasm-component-ld@0.1.5 --root $(BUILD_PREFIX)
 	touch build/wasm-component-ld.BUILT
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ default: build
 	@echo "Use -fdebug-prefix-map=$(ROOT_DIR)=wasisdk://v$(VERSION)"
 
 check:
-	TARGETS="$(TARGETS)" tests/run.sh "$(BUILD_PREFIX)" "$(RUNTIME)" "$(ADAPTER)" "$(WASM_TOOLS)"
+	TARGETS="$(TARGETS)" tests/run.sh "$(BUILD_PREFIX)" "$(RUNTIME)"
 
 clean:
 	rm -rf build $(DESTDIR)
@@ -122,6 +122,11 @@ build/llvm.BUILT:
 		llvm-config
 	touch build/llvm.BUILT
 
+build/wasm-component-ld.BUILT: build/llvm.BUILT
+	cargo install wasm-component-ld@0.1.5 --root $(BUILD_PREFIX)
+	touch build/wasm-component-ld.BUILT
+
+
 # Flags for running `make` in wasi-libc
 # $(1): the target that's being built
 WASI_LIBC_MAKEFLAGS = \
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -122,6 +122,11 @@ build/llvm.BUILT:
 		llvm-config
 	touch build/llvm.BUILT
 
+build/wasm-component-ld.BUILT: build/llvm.BUILT
+	cargo install wasm-component-ld@0.1.5 --root $(BUILD_PREFIX)
```
sbc100, line 126: Most of the other dependencies are expected to be install before running the Makefile (e.g. in the github actions workflow).   Are we sure we want to diverge for this case?

alexcrichton: Sorry I'm not quite sure what you mean, can you clarify? Are you asking whether to submodule/vendor the source of `wasm-component-ld` in this repository? Or otherwise explicitly check that Cargo is itself installed? Or something else perhaps?

I'll also note that `wasm-component-ld` itself has transitive dependencies, so a git submodule wouldn't vendor all the source here. It's possible to vendor all the source, but that seemed pretty heavyweight for just this.

sbc100: Sorry, maybe I wasn't clear.

My point was that the status quo that we don't currently download or install tools when running `make`.    The `Makefile` assumes that all the needed tools already exist in the PATH.

So maybe this `cargo install` command could/should go in `.github/workflows/main.yml` where we take care of pulling in other dependencies like wasmtime and clang?

sbc100: I'm certainly not suggesting we need to build the linker from source.

alexcrichton: To make sure I understand, you're thinking that this rule would do something like:

```
cp `which wasm-component-ld` $(BUILD_PREFIX)/bin
```

If so I can certainly add that, but that means that users building their own wasi-sdk locally will possibly have a different version than what CI in this repo builds.

sbc100: Oh I see, I hadn't quick grokked that we are directly shipping a binary to user here.

But yes, I guess that is what I mean.  However, on reconsideration it does seems a little odd that we are shipping pre-built binaries as part of the SDK.

@sunfishcode WDYT?   Given that we are building llvm itself from source should we maybe also be building this binary from source too?   Conversely, if we decide its reasonable to simply ship upstream binaries maybe we should just ship clang/llvm from the LLVM releases rather than spending all the time to build it from source?   Especially given that we always just pin to llvm releases?

alexcrichton: Looking at the [latest release of LLVM](https://github.com/llvm/llvm-project/releases/tag/llvmorg-18.1.2) it says:

> We suggest you use the binaries from your distribution or build your own if you rely on a specific platform or configuration.

It also looks like there's no release binaries for x64 linux or macOS at all?

Even if there were precompiled binaries, I think that we'd still have the problem of the sysroot not being pre-configured nor the default target being set.

Although don't get me wrong, if LLVM could avoid being built that would make CI and contributing here much less painful

sunfishcode: I agree with @alexcrichton, switching to official LLVM builds is tempting, but they very often don't have binaries for even all the major platforms. And it is convenient to have builds with a default sysroot configured in.

As I understand it, the PR here isn't shipping a pre-built binary to users. `cargo install wasm-component-ld@0.1.5` builds from source. The source comes from crates.io, rather than this repo or a submodule, so we might want to see if we can make it easier for people to build their own wasm-component-ld from source, but overall the approach seems ok to me.

sbc100: Ah I see, I didn't realize this was a build from source.

Can we add a comment for the non-rust folks like me?  Something like "Use cargo to build wasm-component-ld binary from rust source code".

alexcrichton: Certainly!

alexcrichton:
@sbc100 or @sunfishcode did y'all have further thoughts on this?

sbc100:
I'm little wary about this new linker which is outside of llvm and written in rust, but that could also be my lack of familiarity with it.  I think it would be great if it could one day be part of llvm or wasm-ld itself.   Since I have not really been involved with the preview2 or component model stuff myself so I'll leave to you all to decide what is best here.  I'll defer to @sunfishcode for the final lgtm.

alexcrichton:
ping @sunfishcode, do you have thoughts on this?

PR #409, opened 2024-04-12
Don't add git hash to version if minor==0

alexcrichton:
This commit fixes a minor mistake from #392 where the previous perl script had a special case that if the "minor" version was zero then the git hash information wouldn't be printed.

### DIFF
--- a/version.py
+++ b/version.py
@@ -39,7 +39,7 @@ def parse_git_version(version):
         next = parts.pop(0)
         if next == 'm':
             dirty = True
-        else:
+        elif minor != '0':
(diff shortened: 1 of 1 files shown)


abrown (approved):
Thanks for the fix! I should mention here for posterity, since I currently have this paged in: what I called "minor" here is actually not well-named, since I believe what `git describe` drops into this spot is actually "number of commits since last tag." I guess we'll run into issues if we ever decide to do an actual minor release like `wasi-sdk-x.y`; for now at least this will emulate what we have been doing.

Issue #408, opened 2024-04-12
Next release of wasi-sdk-22


alexcrichton:
I wanted to open this up and see if anyone had any objections to publishing a v22 release for wasi-sdk. I'm happy to drive using the instructions @abrown [has added](https://github.com/WebAssembly/wasi-sdk/blob/main/RELEASING.md), but wanted to make sure beforehand there were no objections.

sunfishcode:
Sounds good to me.

alexcrichton:
I've [created a tag](https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-22) which has [kicked off a build](https://github.com/WebAssembly/wasi-sdk/actions/runs/8663908594)

alexcrichton:
The release artifacts are not built correctly so I've deleted the tag.

One issue is https://github.com/WebAssembly/wasi-sdk/pull/409 where `./version.py` was putting a git hash in the version number when it shouldn't. Another issue is CI built artifacts with a version number of 21.21... which should have been 22.0... and I am trying to figure out why and am being pretty unsuccessful.

alexcrichton:
I believe the second issue is https://github.com/WebAssembly/wasi-sdk/pull/410, I'll retry the release once those are merged

abrown:
> One issue is #409 where `./version.py` was putting a git hash in the version number when it shouldn't. Another issue is CI built artifacts with a version number of 21.21... which should have been 22.0... and I am trying to figure out why and am being pretty unsuccessful.

Did my comment in #409 clear things up?

alexcrichton:
Yeah sounds good! I'll try a re-tag now and see how it goes

alexcrichton:
Ok https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-22 was produced just now and things look better. If other folks want to give some eyes on that I can hit the publish button later today or tomorrow.

alexcrichton:
Ok I've hit publish.

abrown:
I opened https://github.com/WebAssembly/wasi-sdk/pull/414 to see if I could figure out where the extra `m` suffix came from.

PR #410, opened 2024-04-12
Force-fetch tags in CI to fix `git describe`

alexcrichton:
My release attempt in #408 did not go well because the release artifacts were not named correctly. This fixes an issue described by actions/checkout#290 where checkouts of annotated tags overwrite the annotation which breaks `git describe`. That means that version detection is broken in CI during releases which causes artifacts to have the wrong information.

This applies the workaround described in that issue to `git fetch --tags --force` after the checkout step to undo the overwrite done in the checkout step.


abrown (approved):
Good find!

PR #412, opened 2024-04-15
Remove cmake_minimum_required from toolchain files

thewtex:
This command is generally not included in toolchain files. It is used in
project configuration files that will consume the toolchain file. Also,
such a low value triggers warnings, support for <3.5.0 is being removed
from CMake.

### DIFF
--- a/wasi-sdk-pthread.cmake
+++ b/wasi-sdk-pthread.cmake
@@ -1,8 +1,5 @@
 # Cmake toolchain description file for the Makefile
 
-# This is arbitrary, AFAIK, for now.
-cmake_minimum_required(VERSION 3.4.0)
-
 set(CMAKE_SYSTEM_NAME WASI)
 set(CMAKE_SYSTEM_VERSION 1)
 set(CMAKE_SYSTEM_PROCESSOR wasm32)
--- a/wasi-sdk-p2.cmake
+++ b/wasi-sdk-p2.cmake
@@ -1,8 +1,5 @@
 # Cmake toolchain description file for the Makefile
 
-# This is arbitrary, AFAIK, for now.
-cmake_minimum_required(VERSION 3.5.0)
-
 # Until Platform/WASI.cmake is upstream we need to inject the path to it
 # into CMAKE_MODULE_PATH.
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}")
--- a/wasi-sdk.cmake
+++ b/wasi-sdk.cmake
@@ -1,8 +1,5 @@
 # Cmake toolchain description file for the Makefile
 
-# This is arbitrary, AFAIK, for now.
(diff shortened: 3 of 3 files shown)


wasi-sdk-pthread.cmake
```diff
@@ -1,8 +1,5 @@
 # Cmake toolchain description file for the Makefile
 
-# This is arbitrary, AFAIK, for now.
-cmake_minimum_required(VERSION 3.4.0)
```
sbc100, line 4: The android NDK toolchain file does seem to include a `cmake_minimum_version`: https://android.googlesource.com/platform/ndk/+/master/build/cmake/android.toolchain.cmake#35

Are you sure this is not supposed to be here?

Perhaps we could bump to to 3.6 like the android one to avoid the warning you mention?

thewtex: Yes, it is not supposed to be there. Was it originally included for a specific reason?

See the [examples in the official CMake documentation](https://github.com/Kitware/CMake/blob/e449417f9468ffada6c0e38a5dab09122a3d1c25/Help/manual/cmake-toolchains.7.rst#toolchain-features). Or [these examples](https://github.com/dockcross/dockcross).

The warning I pointed to is one example of unintended consequences, but there are more. CMake changes its behavior [based on policies](https://cmake.org/cmake/help/latest/command/cmake_policy.html), which are impacted by a `cmake_minimum_required` call.

sbc100: That line as been there since the initial commit (c80964d8d7cf8ca2480d8332cd8022cb0600d421).  

I don't know it was included, it looks like we would have to go back to `yury's wasmception` to find out why.

It sounds like you know what you are talking about cmake-wise, and also emscripten's cmake does not specify `cmake_minimum_required` so I'm tempted to just approve this.

Lets wait for @pchickey to chime in though.

thewtex: Thank you for your attention and consideration.

yamt (approved):

wasi-sdk-pthread.cmake
whitequark: I agree that this change is likely the right one to make.

whitequark (approved):

abrown (approved):
Ok, let's go with this. I'm not as familiar with CMake conventions as I would like but it seems like there is some consensus to remove this. I can see how a user of an already-built toolchain using these files doesn't really care what version of CMake was used to compile the toolchain. And CMake policy [CMP0000](https://cmake.org/cmake/help/latest/policy/CMP0000.html#policy:CMP0000) only really talks about requiring minimum versions in `CMakeLists.txt`.

@thewtex, can you remove this line from the `wasi-sdk-p2.cmake` file as well?

thewtex:
> @thewtex, can you remove this line from the wasi-sdk-p2.cmake file as well?

@abrown 👍 rebased and done.

PR #414, opened 2024-04-16
ci: troubleshoot mingw Git index

abrown:
For some unknown reason, the Windows build of wasi-sdk always creates artifacts that appear as if the current Git index is modified. This troubleshooting commit attempts to figure out why that is; do not merge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -126,6 +126,8 @@ jobs:
           # Upload the dist folder. Give it a name according to the OS it was built for.
           name: ${{ format( 'dist-windows-latest-{0}', matrix.arch) }}
           path: dist
+      - name: What does Git see as modified?
+        run: git diff --stat
 
   dockerbuild:
     name: Docker Build


abrown:
@alexcrichton, I saw wasi-sdk-22 had the extra `m` suffix appended so I ran this to see if I could figure out why. (I also see you ran a similar action but saw that only after I pushed this). In poking around, the only thing I could think is that perhaps mingw is using some old version of Git that falls prey to this [old bug](https://stackoverflow.com/questions/16035240) (here's a [reproducer](https://gist.github.com/peritus/936326/f0ac2a2ac7f3981a1dc045a7f58a7f03f80d7658) I believe).

alexcrichton:
The weird part is that `git describe` seems to work sometimes? In https://github.com/WebAssembly/wasi-sdk/pull/413 I added `python3 ./version.py` before and after the build and it omitted the `+m` before and after, it was only during the build that the `+m` showed up. Given that the `$(shell ...)` to get the version is practically at the top of the `Makefile` I was having a hard time seeing how it could tamper with the tree or change things relative to before-and-after.

Unless... do you think that the `git` used during the github actions step is different than the one used during `make`?

abrown:
> The weird part is that `git describe` seems to work sometimes? In #413 I added `python3 ./version.py` before and after the build and it omitted the `+m` before and after, it was only during the build that the `+m` showed up. Given that the `$(shell ...)` to get the version is practically at the top of the `Makefile` I was having a hard time seeing how it could tamper with the tree or change things relative to before-and-after.
> 
> Unless... do you think that the `git` used during the github actions step is different than the one used during `make`?

Yeah. I'm confused as well. This latest commit ended up with ([ci](https://github.com/WebAssembly/wasi-sdk/actions/runs/8707823201/job/23883738401?pr=414):

- `+m` in the built artifacts
```console
$ find dist
dist
dist/libclang_rt.builtins-wasm32-wasi-22.6gb01b4ded43b5+m.tar.gz
dist/wasi-sdk-22.6gb01b4ded43b5+m-mingw.tar.gz
dist/wasi-sysroot-22.6gb01b4ded43b5+m.tar.gz
```

- no `+m` in either kind of shell (`pwsh` or `bash`):
```console
$ git status
HEAD detached at pull/414/merge
nothing to commit, working tree clean
$ git diff --stat
$ python3 version.py
22.6gb01b4ded43b5
```

PR #415, opened 2024-04-23
Add python3 as build requirement to README

imrying:
When trying to build the wasi-sdk I noticed that `python3` was not mentioned in the build process requirements. Since it is not part of the standard environment on most distributions I thought it would be a good idea to add it.
Cheers!


sbc100:
Ah yes that was very recently added in 2436b99f1d59becb2d2e6b3e11dd6e48903a6adb

PR #425, opened 2024-06-08
add Linux/ARM64 cross-compilation support

### LINKED ISSUE #236: Add binary release for Linux on Apple Silicon M1

### LINKED ISSUE #347: Package release for Linux ARM64
I'm building directly on an arm64 linux machine and compiling llvm takes 60+ minutes so would be really helpful to have a release build. 

Is there currently any plan for this? I see a similar, but apple-related issue here #236 as well as some mention of failed arm builds on linux in other tickets. Would be really helpful.

Cheers

dicej:
Currently, the Makefile assumes the LLVM toolchain it builds can be executed natively to build `wasi-libc` etc., which isn't true when cross-compiling for another platform, but we can work around that by:

1. Building the native LLVM toolchain and using it to build everything else, as usual
2. Deleting that LLVM build and rebuilding (and reinstalling) it with `LLVM_CMAKE_FLAGS` set to cross compile
3. Rebuilding and reinstalling a cross-compiled `wasm-component-ld`
4. Building deb and tar files from the above

Note that we now label the tarfiles `linux-amd64` and `linux-arm64`, respectively for clarity.

The whole approach is a bit hacky, but GitHub is planning to roll out ARM64 runner support for open source projects later this year, at which point we can start building natively, so I don't think we need to invest a lot of effort into this.

I've run CI in my fork and verified the artifact produced there works on my Ubuntu 24.04 ARM64 machine (Asahi Linux on an Apple M2 Pro).

~Note that I have not yet updated the `dockerbuild` CI step, so the `dist-ubuntu-bionic` artifacts do not yet include an ARM64 build; I'll experiment with that next.~  Update: now we _only_ do a cross build as part of `dockerbuild`, since `dist-ubuntu-bionic` is the one we actually publish in releases.

Fixes #236
Fixes #347

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,7 +60,7 @@ jobs:
         run: sudo apt install ccache ninja-build
         if: matrix.os == 'ubuntu-latest'
       - name: Build
-        run: NINJA_FLAGS=-v make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON
+        run: NINJA_FLAGS=-v make package LLVM_CMAKE_FLAGS=-DLLVM_CCACHE_BUILD=ON DEB_ARCH=amd64 TAR_MACHINE=linux-amd64
         shell: bash
       - name: Run the testsuite
         run: NINJA_FLAGS=-v make check RUNTIME=wasmtime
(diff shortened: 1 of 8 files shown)


dicej:
Oops, the new step shouldn't be run on macOS, obviously.  I'll fix that.

dicej:
Yeah, that all sounds great to me, and I 100% agree that there's a lot of duplicated and wasted effort, including the redundant builds of the sysroot and the `ubuntu-latest` and 32-bit Windows artifacts which we build but then discard in favor of the `ubuntu-bionic` and 64-bit Windows ones.  Trimming that bloat would be reason enough to refactor this thing, not to mention the maintenance benefits.

OTOH, I don't have the bandwidth right now for a significant refactor.  Happy to open an issue to track it, though!

dicej:
I just pushed an update; now we _only_ build for Linux/ARM64 as part of the `dockerbuild` job since `dist-ubuntu-bionic` is the build we actually care about.

alexcrichton:
I'm personally always very hesitant to put stop energy on things when I don't have a great alternative. To that effect I've tried to put my money where my mouth is in https://github.com/WebAssembly/wasi-sdk/pull/429

alexcrichton:
I believe this is handled now in https://github.com/WebAssembly/wasi-sdk/pull/429. The [latest CI build](https://github.com/WebAssembly/wasi-sdk/actions/runs/9910478645) has various artifacts and the "release-artifacts" artifact contains all the architectures. It's a bit big but downloading that shows it has `wasi-sdk-23.3g2e7bb65533b2-arm64-linux.tar.gz` and `wasi-sdk-23.3g2e7bb65533b2-arm64-linux.deb` which I believe are the desired artifacts. That means these should all show up on the next release.

PR #430, opened 2024-06-18
add wasi-sdk-p2.cmake

dicej:
This is analogous to wasi-sdk.cmake and wasi-sdk-pthread.cmake, but for WASIp2.

### DIFF
--- a/wasi-sdk-p2.cmake
+++ b/wasi-sdk-p2.cmake
@@ -1,0 +1,40 @@
+# Cmake toolchain description file for the Makefile
+
+# This is arbitrary, AFAIK, for now.
+cmake_minimum_required(VERSION 3.4.0)
+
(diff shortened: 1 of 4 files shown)


wasi-sdk-p2.cmake
```diff
@@ -0,0 +1,40 @@
+# Cmake toolchain description file for the Makefile
```
akoeplinger, line 1: the file needs to be moved into the `docker` folder

dicej: Ah, thanks for pointing that out.  TIL there are two versions of each of the wasi-sdk*.cmake files.

alexcrichton (approved):
FWIW I've fixed the duplication issue at https://github.com/WebAssembly/wasi-sdk/pull/429, although there's still duplication between each of the `wasi-*.cmake` files, but that seems more minor.

PR #424, opened 2024-06-07
Add full LTO build of wasi-libc and libc++

yamt:

PR #436, opened 2024-07-12
Add full LTO build of wasi-libc and libc++

yamt:
redo https://github.com/WebAssembly/wasi-sdk/pull/424 after cmake migration

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -160,7 +160,12 @@ endforeach()
 # libcxx build logic
 # =============================================================================
 
-function(define_libcxx target)
+execute_process(
+  COMMAND ${CMAKE_C_COMPILER} -dumpversion
+  OUTPUT_VARIABLE llvm_version
+  OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+function(define_libcxx_sub target target_suffix extra_target_flags extra_libdir_suffix)
   if(${target} MATCHES threads)
     set(threads ON)
     set(pic OFF)
@@ -170,6 +175,10 @@ function(define_libcxx target)
     set(pic ON)
     set(target_flags "")
   endif()
+  if(${target_suffix} MATCHES lto)
+    set(pic OFF)
+  endif()
+  list(APPEND target_flags ${extra_target_flags})
 
   set(runtimes "libcxx;libcxxabi")
 
@@ -188,7 +197,7 @@ function(define_libcxx target)
   set(extra_cxxflags_list ${CMAKE_CXX_FLAGS} ${extra_flags})
   list(JOIN extra_cxxflags_list " " extra_cxxflags)
 
-  ExternalProject_Add(libcxx-${target}-build
+  ExternalProject_Add(libcxx-${target}${target_suffix}-build
     SOURCE_DIR ${llvm_proj_dir}/runtimes
     CMAKE_ARGS
       ${default_cmake_args}
@@ -226,8 +235,8 @@ function(define_libcxx target)
       -DUNIX:BOOL=ON
       -DCMAKE_C_FLAGS=${extra_cflags}
       -DCMAKE_CXX_FLAGS=${extra_cxxflags}
-      -DLIBCXX_LIBDIR_SUFFIX=/${target}
-      -DLIBCXXABI_LIBDIR_SUFFIX=/${target}
+      -DLIBCXX_LIBDIR_SUFFIX=/${target}${extra_libdir_suffix}
+      -DLIBCXXABI_LIBDIR_SUFFIX=/${target}${extra_libdir_suffix}
 
     # See https://www.scivision.dev/cmake-externalproject-list-arguments/ for
     # why this is in `CMAKE_CACHE_ARGS` instead of above
@@ -241,6 +250,11 @@ function(define_libcxx target)
     USES_TERMINAL_BUILD ON
     USES_TERMINAL_INSTALL ON
   )
+endfunction()
+
+function(define_libcxx target)
+  define_libcxx_sub(${target} "" "" "")
+  define_libcxx_sub(${target} "-lto" "-flto=full" "/llvm-lto/${llvm_version}")
 
   # As of this writing, `clang++` will ignore the target-specific include dirs
   # unless this one also exists:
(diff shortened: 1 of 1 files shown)


cmake/wasi-sdk-sysroot.cmake
```diff
@@ -216,14 +237,19 @@ function(define_libcxx target)
     USES_TERMINAL_BUILD ON
     USES_TERMINAL_INSTALL ON
   )
+endfunction()
+
+function(define_libcxx target)
+  define_libcxx_sub(${target} "" "" "")
+  define_libcxx_sub(${target} "-lto" "-flto=full" "/llvm-lto/${llvm_version}")
```
alexcrichton, line 244: Is this `/llvm-lto/$version` convention something codified by Clang itself? Or something we're making up here in wasi-sdk?

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -124,8 +127,17 @@ function(define_wasi_libc target)
...
         ${extra_make_flags}
+    COMMAND
+      ${MAKE} -j8 -C ${build_dir}
+        CC=${CMAKE_C_COMPILER}
+        AR=${CMAKE_AR}
+        NM=${CMAKE_NM}
+        SYSROOT=${wasi_sysroot}
+        EXTRA_CFLAGS=${extra_cflags}
+        TARGET_TRIPLE=${target}
+        ${extra_make_flags_lto}
```
alexcrichton, line 140: Could this be split into a separate build rule with a separate build directory? I think it'd be reasonable to add an option to disable/enable the LTO builds and having it bundled in here will make it difficult to separate out. For debugging it would also be useful to have a single build target for just LTO or just without LTO.

alexcrichton (commented):
As a high-level question, does Clang support embedding LLVM IR in object files? That's, for example, how the Rust compiler works without having two entirely different builds of the standard library. The single precompiled binary for the standard library supports both LTO and non-LTO builds.

sbc100:
> As a high-level question, does Clang support embedding LLVM IR in object files? That's, for example, how the Rust compiler works without having two entirely different builds of the standard library. The single precompiled binary for the standard library supports both LTO and non-LTO builds.

I've never heard of that before.  I've only ever seen separate object files/libraries used.   I'm very curious to learn about this feature!

cmake/wasi-sdk-sysroot.cmake
sbc100: This is something that was created/added specifically for the wasm clang driver: https://github.com/llvm/llvm-project/blob/0309709a6786653da7164334c83b09c9f37b943a/clang/lib/Driver/ToolChains/WebAssembly.cpp#L206-L212

See https://reviews.llvm.org/D70677

yamt:
> As a high-level question, does Clang support embedding LLVM IR in object files? That's, for example, how the Rust compiler works without having two entirely different builds of the standard library. The single precompiled binary for the standard library supports both LTO and non-LTO builds.

llvm has a concept called "fat lto object".
but it doesn't support wasm right now.

cmake/wasi-sdk-sysroot.cmake
alexcrichton: Ah ok! Could that be documented here that the convention is derived from LLVM itself?

alexcrichton:
The Rust compiler at least has a flag to embed bitcode in the original object file itself. It generates LLVM IR, uses that to generate an object file, then encodes the LLVM IR and puts that in the object file as well. Then when LTO is performed the compiler will recognize this type of file, extract the LLVM IR, and then process it as if it were LTO. The original object never gets passed to the linker in LTO builds.

LTO works relatively differently in Rust than it does in C/C++ though. Contrary to the name it's all done within rustc itself, rustc by default doesn't rely on the linker to perform LTO. Whether or not that would work here I'm not sure, but given that there's only a single linker in the wasm ecosystem and compiler it would be pretty nice to avoid two entirely separate builds of wasi-libc/etc to be configured just right to get LTO.

alexcrichton:
To be clear I wasn't saying this should all be held up to do something toally different, I was curious if something was possible. Sounds like it isn't. I think this is fine to land modulo a few small comments from myself above.

cmake/wasi-sdk-sysroot.cmake
yamt: do you mean to make 4 more copies of wasi-libc source tree?

alexcrichton: Until wasi-libc supports out-of-tree builds yes I think that will be required.

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -143,7 +155,12 @@ endforeach()
 # libcxx build logic
 # =============================================================================
 
-function(define_libcxx target)
+execute_process(
+  COMMAND ${CMAKE_C_COMPILER} -dumpversion
+  OUTPUT_VARIABLE llvm_version
```
alexcrichton, line 160: Can this also be called something like `host_llvm_version` or `clang_llvm_version` to better distinguish from [`clang_version`](https://github.com/WebAssembly/wasi-sdk/blob/de6328701c9bead32fdbcabf8b177fc0c8c91011/CMakeLists.txt#L29)? (either that or can that variable be renamed?)

yamt: my impression is clang_version should be retired. we can use -print-resource-dir instead.

cmake/wasi-sdk-sysroot.cmake
yamt: done

cmake/wasi-sdk-sysroot.cmake
yamt: i added a comment

cmake/wasi-sdk-sysroot.cmake
alexcrichton: Can you retire it in this PR then? Otherwise this is creating a confusing state where both are available and they refer to slightly different things.

yamt: i don't want to interfere with https://github.com/WebAssembly/wasi-sdk/pull/445.

alexcrichton: Now that #445 is merged could this be retired?

alexcrichton:
Also one thing I just remembered as well, could a test be added for this? Or also I'm not sure how the existing tests interact with this because they're already passing `-flto` and passing, so I'm not sure how this relates to that.

yamt:
> Also one thing I just remembered as well, could a test be added for this? Or also I'm not sure how the existing tests interact with this because they're already passing `-flto` and passing, so I'm not sure how this relates to that.

the existing tests using -flto somehow cover this. i have already fixed a bug to pass them.

cmake/wasi-sdk-sysroot.cmake
yamt: see https://github.com/WebAssembly/wasi-sdk/pull/451

yamt:
the last push is a conflict resolution.

alexcrichton (approved):

PR #437, opened 2024-07-12
README.md: update build instructions

yamt:


alexcrichton (approved):
Thanks! I'll follow up with some more docs as well.

PR #445, opened 2024-07-15
Don't modify the host compiler's sysroot

### LINKED ISSUE #444: Support building just a sysroot, not a toolchain
Right now this is sort of supported with the default CMake invocation of this repository but as pointed out in [this comment](https://github.com/WebAssembly/wasi-sdk/pull/443#issuecomment-2226991677) and the discussion in that thread this doesn't fully work because compiler-rt is copied into the host toolchain. That's generally not a great thing to do.

Fixing this would require one of two things that I can think of:

1. Support not building compiler-rt. How exactly the host toolchain gets compiler-rt, however, I'm not certain.
2. Support the host compiler using the just-built compiler-rt. It sounds like `-resource-dir` is the way to go here according to the discussion in #443 but I don't know how to communicate that or how best to document that (e.g. what is the exact structure of `-resource-dir`? What to do with target-specific headers? etc)

alexcrichton:
This commit updates the building of the wasi-sdk sysroot to leverage the `-resource-dir` argument from Clang to avoid modifying the host compiler's sysroot with compiler-rt things. This should help improve the experience of building a standalone sysroot with whatever host Clang is on the system.

Closes #444

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -12,12 +12,17 @@ option(WASI_SDK_INCLUDE_TESTS "Whether or not to build tests by default" OFF)
 
 set(wasi_tmp_install ${CMAKE_CURRENT_BINARY_DIR}/install)
 set(wasi_sysroot ${wasi_tmp_install}/share/wasi-sysroot)
+set(wasi_resource_dir ${wasi_tmp_install}/lib/clang/${clang_version})
+
+# Force usage of the custom-built resource-dir and sysroot for the rest of the
+# wasi compiles.
+add_compile_options(-resource-dir ${wasi_resource_dir})
+add_compile_options(--sysroot ${wasi_sysroot})
 
 if(WASI_SDK_DEBUG_PREFIX_MAP)
   add_compile_options(
     -fdebug-prefix-map=${CMAKE_CURRENT_SOURCE_DIR}=wasisdk://v${wasi_sdk_version})
 endif()
-
 
 # Default arguments for builds of cmake projects (mostly LLVM-based) to forward
 # along much of our own configuration into these projects.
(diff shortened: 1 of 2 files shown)


abrown:
cc: @whitequark, does this help towards resolving the discussion you were having over in https://github.com/WebAssembly/wasi-sdk/pull/443#issuecomment-2229049509? I'm thinking of merging this but wanted to hear whether you had any additional feedback here.

whitequark (approved):
LGTM

abrown (approved):

yamt:
have you ever tested if this change actually "avoid modifying the host compiler's sysroot with compiler-rt things"?
i don't think it works w/o https://github.com/WebAssembly/wasi-libc/pull/517.
i suspect we need a bit more CI coverage.

alexcrichton:
Yes I tested this doesn't modify the host compiler's sysroot. No there is no test of using the host toolchain in CI. Yes it would be great to add tests.

yamt:
> Yes I tested this doesn't modify the host compiler's sysroot.

did your "host compiler" already happen to have wasi compiler-rt?
otherwise, i don't understand why you didn't need https://github.com/WebAssembly/wasi-libc/pull/517.
for me, gmake tried to create the directory for wasi compiler-rt in the host compiler's resource dir.

alexcrichton:
On Ubuntu 24.04 I ran:

```
$ cmake -G Ninja -B build -S . -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_NM=llvm-nm-18
$ ninja -C build dist
```

and everything compiled ok. No the host compiler does not have a compiler-rt for wasm.

alexcrichton:
The `clang` package owns clang/clang++ and `llvm-18` owns `llvm-nm`. Yes, they're from ubuntu packages.

As to what links compiler-rt is linked, I'm not sure.

PR #446, opened 2024-07-15
Update build of toolchain/sysroot to not touch installation root

### LINKED ISSUE #442: Running `make` in a configured wasi-sdk tree installs things into `/usr/local`
I did not expect this behavior (normally only `make install` does that, which I was going to run with `DESTDIR=`, as usual) and it is fairly unwelcome in that I don't know how or if I can uninstall those files.

alexcrichton:
This changes everything to ensure that only the `install` step actually tries to install things. Everything is staged into temporary `./install` folders inside of the build directory and then running the build system's `install` target will actually copy out everything using CMake builtins.

Closes #442


sbc100 (approved):
LGTM!

PR #451, opened 2024-07-18
Retire clang_version

yamt:
Note: this makes the "install" target install the compiler rt into the compiler's runtime directory.  IMO, it's what "install" is supposed to do.  If you want to avoid modifing the runtime directory for some reasons, you can still do "dist" without "install".

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -12,7 +12,7 @@ option(WASI_SDK_INCLUDE_TESTS "Whether or not to build tests by default" OFF)
 
 set(wasi_tmp_install ${CMAKE_CURRENT_BINARY_DIR}/install)
 set(wasi_sysroot ${wasi_tmp_install}/share/wasi-sysroot)
-set(wasi_resource_dir ${wasi_tmp_install}/lib/clang/${clang_version})
+set(wasi_resource_dir ${wasi_tmp_install}/wasi-resource-dir)
 
 # Force usage of the custom-built resource-dir and sysroot for the rest of the
 # wasi compiles.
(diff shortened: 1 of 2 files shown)


alexcrichton:
Personally I think that "install" should only ever modify things under `CMAKE_INSTALL_PREFIX` by default. Having an option to modify the host compiler's sysroot seems reasonable, but I don't think by default this should modify the host compiler's sysroot.

yamt:
> Personally I think that "install" should only ever modify things under `CMAKE_INSTALL_PREFIX` by default. Having an option to modify the host compiler's sysroot seems reasonable, but I don't think by default this should modify the host compiler's sysroot.

i tend to disagree. but i pushed a change to avoid modifying the host compiler's resource dir.

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -243,9 +243,17 @@ endforeach()
...
         USE_SOURCE_PERMISSIONS
         DESTINATION ${CMAKE_INSTALL_PREFIX})
+cmake_path(IS_PREFIX CMAKE_INSTALL_PREFIX ${clang_resource_dir} NORMALIZE install_resource_dir)
+if(install_resource_dir)
+  install(DIRECTORY ${wasi_resource_dir}/lib
+          USE_SOURCE_PERMISSIONS
+          DESTINATION ${clang_resource_dir})
+else()
+  message(STATUS "The resource dir (${clang_resource_dir}) will not be updated by the install target because it's out of CMAKE_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX}")
+endif()
```
alexcrichton, line 256: This to me seems odd to me to have a conditional installation. Could an option be added to modify the host compiler's resource-dir and if that's turned off then the preexisting install location is used? (or something similar to it, probalby not the exact same structure with a clang version that isn't guaranteed to match something)

yamt: i'm a bit confused because i just tried to follow your suggestion and you are now saying it's odd. :-)

how about the last push?

alexcrichton: The first version of this PR unconditionally modified the host compiler resource-dir which I felt was surprising since it was modifying things outside of `CMAKE_INSTALL_PREFIX`. The second version of this PR, which this comment was targetting, did nothing by default and happened to install things if `CMAKE_INSTALL_PREFIX` lined up. I found that surprising as well because it's not documented and just a status message can be easily lost in configure output.

Third version of this PR makes more sense. Something is always installed and it always shows up somewhere, and an opt-in is enabled to perform a non-standard installation to put something outside of the installation prefix.

alexcrichton (approved):

Issue #450, opened 2024-07-17
error[E0658]: using an imported function as entry point `main` is experimental


yamt:
i saw the following error while doing "cmake --build build/toolchain --target dist".
i guess rust/cargo requirements should be documented in https://github.com/WebAssembly/wasi-sdk?tab=readme-ov-file#requirements.

```shell
error[E0658]: using an imported function as entry point `main` is experimental
 --> /Users/yamamoto/.cargo/registry/src/index.crates.io-6f17d22bba15001f/wasm-component-ld-0.5.5/src/main.rs:1:5
  |
1 | use wasm_component_ld::main;
  |     ^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: see issue #28937 <https://github.com/rust-lang/rust/issues/28937> for more information

For more information about this error, try `rustc --explain E0658`.
error: could not compile `wasm-component-ld` (bin "wasm-component-ld") due to 1 previous error
error: failed to compile `wasm-component-ld v0.5.5`, intermediate artifacts can be found at `/var/folders/74/hw1sphgx0lv63q6pq_n5grw00000gn/T/cargo-installJIGQGF`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
ninja: build stopped: subcommand failed.
```

alexcrichton:
Yes this is a toolchain error where the Rust toolchain you're using is too old. Once https://github.com/bytecodealliance/wasm-component-ld/pull/36 gets published you'll get a better first-class error message when the toolchain is too old.

alexcrichton:
I believe this has since been fixed, so closing.

PR #452, opened 2024-07-18
README.md: document rust requirement

yamt:
Fixes https://github.com/WebAssembly/wasi-sdk/issues/450

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ The Wasm-sdk's build process needs some packages :
 * `clang`
 * `ninja`
 * `python3`
+* The latest rust toolchain
 
 Please refer to your OS documentation to install those packages.
 


yamt:
* i myself is not entirely happy with the rust requirement. but it's the current reality.
* i don't know how latest it needs to be. rustc 1.77.0 didn't work. rustc 1.79.0 seems ok.

README.md
```diff
@@ -44,6 +44,7 @@ The Wasm-sdk's build process needs some packages :
 * `clang`
 * `ninja`
 * `python3`
+* The latest rust toolchain
```
abrown, line 47: Looks like [this](https://github.com/bytecodealliance/wasm-component-ld/pull/36/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R9) would indicate that 1.76.0 should work? But you mention that 1.77.0 did not work for you... what did you see?

yamt: see https://github.com/WebAssembly/wasi-sdk/issues/450

yamt: @abrown do you have any more concerns?

abrown: Well, I kind of was thinking we could be more precise here and since `wasm-component-ld` is being successfully with Rust 1.76.0 in CI we should probably document that. But then I'm confused why 1.77.0 didn't work for you? Or was something else the problem?

yamt: i dunno. i merely blindly followed alex's advice to update the toolchain.

but is it worth to block this PR for a month while we don't document such precise verisions for other requirements either? (cmake python etc)

alexcrichton: I don't think it's worth merging an inaccurate claim such as this because it's not true that "the latest rust toolchain" is required to build wasm-component-ld. I think it's better to either rename this to `rustc`, as the other tools are here, or work towards a stricter requirement here such as [updating the project to test a minimum version](https://github.com/bytecodealliance/wasm-component-ld/pull/36) and then listing that version here.

abrown: I sure wasn't trying to block any PRs, it's just that something here didn't make sense so I moved on to other things. Now that I've circled back, I think since it isn't clear to you which version you used but CI would indicate that the minimum supported version is 1.76.0, we can document that. That way we can avoid telling users to do extra toolchain updates; it sounds like the extra overhead of new toolchain updates is something you wouldn't be happy with.

abrown, line 47:
```suggestion
* [Rust](https://rustup.rs) v1.76.0 or later
```

abrown: [oops, commented on a stale page and didn't see @alexcrichton's reply; I suggested an update that should resolve this]

yamt: > I don't think it's worth merging an inaccurate claim such as this because it's not true that "the latest rust toolchain" is required to build wasm-component-ld.

Sorry, which part of the claim was incorrect?
"Latest"?
1.77.0 didn't work and you adviced me I was using too old toolchian. So I updated it to 1.79.0 and it worked.
Isn't it inappropriate to call 1.79.0 latest?

yamt: > since it isn't clear to you which version you used

It's clear it was 1.77.0.

abrown: Ok, I guess I was the one who misunderstood, sorry! But something still doesn't seem quite right with what you reported. I tried to replicate what you found and I'm seeing that `wasm-component-ld` builds just fine with version v1.76.0:

```console
$ cargo install 1.76.0
$ cargo default 1.76.0
$ cargo version
cargo 1.76.0 (c84b36747 2024-01-18)
$ cmake -G Ninja -B build/toolchain -S . -DWASI_SDK_BUILD_TOOLCHAIN=ON -DCMAKE_INSTALL_PREFIX=build/install
...
$ cmake --build build/toolchain --target wasm-component-ld
...
   Compiling wasm-component-ld v0.5.6
    Finished release [optimized] target(s) in 18.50s
  Installing /.../wasi-sdk/build/toolchain/wasm-component-ld/bin/wasm-component-ld
   Installed package `wasm-component-ld v0.5.6` (executable `wasm-component-ld`)
```

If I run the same with v1.75.0, I see the following failure:

```
error: cannot install package `wasm-component-ld 0.5.6`, it requires rustc 1.76.0 or newer, while the currently active rustc version is 1.75.0
```

Is there any chance something else was wrong with the environment you ran v1.77.0 in?

alexcrichton: > Sorry, which part of the claim was incorrect? "Latest"?

Yes, this is incorrect. For example when you created this PR Rust 1.79 was the latest Rust release. I believe what you were trying to document originally was that Rust 1.79 was required. As of today, however, Rust 1.80 is the latest Rust release. That means that the claim that the "latest" Rust release is needed is not correct.

Additionally `wasm-component-ld` as-used by wasi-sdk requires 1.76.0, and as Andrew pointed out this is tested in CI. This additionally means that the claim that the latest compiler is required is incorrect.

abrown: Maybe this should read "`cargo` and `rustc` v1.76.0 or later" for consistency

yamt: > Ok, I guess I was the one who misunderstood, sorry! But something still doesn't seem quite right with what you reported. I tried to replicate what you found and I'm seeing that `wasm-component-ld` builds just fine with version v1.76.0:
> 
> ```
> $ cargo install 1.76.0
> $ cargo default 1.76.0
> $ cargo version
> cargo 1.76.0 (c84b36747 2024-01-18)
> $ cmake -G Ninja -B build/toolchain -S . -DWASI_SDK_BUILD_TOOLCHAIN=ON -DCMAKE_INSTALL_PREFIX=build/install
> ...
> $ cmake --build build/toolchain --target wasm-component-ld
> ...
>    Compiling wasm-component-ld v0.5.6
>     Finished release [optimized] target(s) in 18.50s
>   Installing /.../wasi-sdk/build/toolchain/wasm-component-ld/bin/wasm-component-ld
>    Installed package `wasm-component-ld v0.5.6` (executable `wasm-component-ld`)
> ```
> 
> If I run the same with v1.75.0, I see the following failure:
> 
> ```
> error: cannot install package `wasm-component-ld 0.5.6`, it requires rustc 1.76.0 or newer, while the currently active rustc version is 1.75.0
> ```
> 
> Is there any chance something else was wrong with the environment you ran v1.77.0 in?

i can still reproduce https://github.com/WebAssembly/wasi-sdk/issues/450 with rustc 1.77.0 + wasm-component-ld 0.5.5.
(we bumped wasm-component-ld version to 0.5.6, which doesn't seem to have the problem, after this PR.)

yamt: > Yes, this is incorrect. For example when you created this PR Rust 1.79 was the latest Rust release. I believe what you were trying to document originally was that Rust 1.79 was required. As of today, however, Rust 1.80 is the latest Rust release. That means that the claim that the "latest" Rust release is needed is not correct.

ok. i dropped the "latest" claim. the main claim of this PR was "we require rust".

PR #455, opened 2024-07-22
Refactor handling env vars in CI build

alexcrichton:
This commit updates the handling of various environment variables to centralize them in one location in `main.yml` as part of CI configuration. This removes a few custom-named variables in favor of using explicitly-listed env vars. Additionally this moves some env var management from the CI container files to the github actions config as well to keep everything in one place ideally.

### DIFF
--- a/ci/docker/Dockerfile.arm64-linux
+++ b/ci/docker/Dockerfile.arm64-linux
@@ -2,22 +2,4 @@ FROM wasi-sdk-builder-base
 
 RUN apt-get install -y g++-aarch64-linux-gnu
 
-# Don't build a sysroot for this cross-compiled target since it would require a
-# host compiler and the sysroot is otherwise already built on other CI builders.
-ENV WASI_SDK_CI_SKIP_SYSROOT 1
-
-ENV WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS \
-  -DCMAKE_C_COMPILER=aarch64-linux-gnu-gcc \
-  -DCMAKE_CXX_COMPILER=aarch64-linux-gnu-g++ \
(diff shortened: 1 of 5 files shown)


ci/docker/Dockerfile.arm64-linux
```diff
@@ -2,22 +2,4 @@ FROM wasi-sdk-builder-base
...
-  -DCMAKE_CROSSCOMPILING=True \
-  -DCMAKE_CXX_FLAGS=-march=armv8-a \
-  -DCMAKE_SYSTEM_PROCESSOR=arm64 \
-  -DCMAKE_SYSTEM_NAME=Linux \
-  -DLLVM_HOST_TRIPLE=aarch64-linux-gnu \
-  -DRUST_TARGET=aarch64-unknown-linux-gnu
-
-ENV WASI_SDK_CI_TOOLCHAIN_CMAKE_ARGS \
-  -DWASI_SDK_ARTIFACT=arm64-linux \
-  -DRUST_TARGET=aarch64-unknown-linux-gnu
```
abrown, line 21: If run in some local workflow, these `Dockerfile`s no longer will produce the expected output (or perhaps even work?). How do you feel about pointing that (potential) user back to `main.yml` for the proper `WASI_SDK_CI_*` that is expected from them?

abrown (approved):

ci/docker/Dockerfile.arm64-linux
alexcrichton: That's right yeah, although that was also already sort of true for the Windows/macOS builds since they've also got configuration in CI that would have to be copied down loacally. I'll leave some notes in the containers though.

PR #456, opened 2024-07-22
Use MSVC for the Windows toolchain instead of MinGW

### LINKED ISSUE #454: wasi-sdk-23 requires libgcc_s_seh-1.dll again
Opening a new issue for https://github.com/WebAssembly/wasi-sdk/issues/148#issuecomment-2240979412 for more visibility. I'll work on investigating this.

alexcrichton:
Explicitly use MSVC to avoid the runtime dependencies that the default toolchain CMake is using is bringing in.

Closes #454

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,6 +110,9 @@ jobs:
           echo WASI_SDK_CI_TOOLCHAIN_CMAKE_ARGS="$cmake_args" >> $GITHUB_ENV
         shell: bash
 
+      - name: Clear ccache statistics
+        run: ccache --zero-stats
+
       # Add some extra installed software on each runner as necessary.
       - name: Setup `wasmtime` for tests
         uses: bytecodealliance/actions/wasmtime/setup@v1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -137,7 +137,7 @@ jobs:
       # MSVC to compile LLVM as that avoids extra runtime dependencies
       # msys/mingw might bring.
       #
-      # As of 2024-07-22 this sha is the "v1" tag.
+      # As of 2024-07-22 this sha is the "v1.13.0" tag.
       - uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756
         if: runner.os == 'Windows'
       - name: Build and test (Windows)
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -118,11 +118,23 @@ jobs:
...
         if: runner.os == 'Linux'
 
-      # Use a shorter build directory than the default on Windows to avoid
-      # hitting path length and command line length limits. See
-      # WebAssembly/wasi-libc#514
+      # Setup the VS Developoer Prompt environment variables to explicitly use
+      # MSVC to compile LLVM as that avoids extra runtime dependencies
+      # msys/mingw might bring.
+      #
+      # As of 2024-07-22 this sha is the "v1" tag.
```
abrown, line 125:
```suggestion
      # As of 2024-07-22 this sha is the "v1.13.0" tag.
```

.github/workflows/main.yml
```diff
@@ -118,11 +118,23 @@ jobs:
...
+      - uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756
```
abrown, line 126: I was debating between the need for this action versus the risk: it looks like ~17k other projects depend on this, it has more than a few contributors, and the action is relatively simple &mdash; if it solves the problem then let's go for it.

abrown (approved):

.github/workflows/main.yml
alexcrichton: Yeah I was wrestling with the same. I spend 30 min or so googling around to see if there were other common solutions to this but everything pointed to a github action one way or another so I opted to do that with a specific sha to at least be insulated against changes.

alexcrichton:
Got a successful CI run, but I'm going to rerun CI to confirm that caching makes it so the next run isn't 2 hours. Afterwards I'll download this on a windows machine and test that things work as expected.

pavelsavara:
I was able to compile dotnet/mono using the [artifact](https://github.com/WebAssembly/wasi-sdk/actions/runs/10046566465/artifacts/1728230093) from this PR on my local windows

alexcrichton:
The "cached build" took 2+ hours so I'm going to try to dig in to why ccache isn't working.

alexcrichton:
Hm on a new commit it worked with a ~20m build time, similar to today. Given that I think this is ok, so I'm going to merge.

.github/workflows/main.yml
sebgod: personally, I would fork any dep like this (in case the repo gets deleted, repurposed, ...)

PR #462, opened 2024-07-26
Add some CMake guards when building the sysroot

alexcrichton:
* Require that the compiler is Clang, for example gcc and msvc cannot compile to WebAssembly.
* Require that the Clang version is above the minimum threshold. Unsure what the minimum threshold is at this time so I've set it to 18.0.0

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -5,6 +5,16 @@ if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE RelWithDebInfo)
 endif()
 
+if(NOT CMAKE_C_COMPILER_ID MATCHES Clang)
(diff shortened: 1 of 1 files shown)


abrown (commented):
Why set the minimum? I.e., if it's too high then old-toolchain users run into issues; if it's too low then what's the point?

alexcrichton:
It's mostly the first part, "users run into issues," where an error message saying "your toolchain is too old" is often more debuggable than some obscure "whatever error this old compiler happened to emit" error message. In my experience in Rust most folks outside of maintainers don't know how to diagnose such errors since the version isn't the obvious culprit

abrown (approved):
Makes sense; let's merge it. I would propose we also keep open the possibility of lowering the minimum in the future: if any users of older toolchains feel strongly about this, maybe we allow them to downgrade this?

PR #463, opened 2024-07-29
Disable docker cache with github actions

alexcrichton:
The container in this repository does not take long to build and the caches look like they're take a nontrivial amount of space in CI. Try to free up more room for LLVM caches which are the main reason to use caching in this repository.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -242,8 +242,6 @@ jobs:
           platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
 
       - name: Publish a draft release
         if: startsWith(github.ref, 'refs/tags')


abrown (approved):
I think this is fine, but wanted to note what it seems to change:
- the `Build and test (Linux)` step calls `docker-build.sh`, which builds everything in Docker
- the `Build x86_64-linux` matrix job calls this step and (unscientifically) was around 21 minutes previous to this change
- this change seems to increase the `Build and test (Linux)` step to 26 minutes, presumably because we need to re-pull some Docker images (?)

In any case, I don't think this is the longest pole in the tent: the `Build x86_64-windows` job is taking 40-50 minutes... wonder why that is so much slower?

alexcrichton:
This should only affect the "Finalize wasi-sdk artifacts" action and shouldn't have any affect on the other jobs. Specifically the "Build and push wasi-sdk docker image" step is affected by this caching. Looking at a previous run it looks like it took 52s and on this PR it took 5m49s. I didn't realize this would lead to such a large discrepancy in performance. My guess is that building the arm64 container image requires a lot to run through QEMU and that probably isn't the fastest thing in the world.

alexcrichton:
Oh, as for windows, looks like while a cache was restored `ccache -s` reported mostly misses, so for whatever reason the restored cache wasn't valid (not sure why myself)

PR #465, opened 2024-07-31
Add a test on CI that just the sysroot can be built

alexcrichton:
This commit adds a test and matrix entry to CI which asserts that the sysroot can be built to run tests using a stock Clang compiler found on the system. This fixes a few minor issues as well in developing this.

This additionally refactors CI a bit to move shared steps amongst jobs into separate composite actions in this repository to avoid duplication across jobs.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,14 +72,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - run: git fetch --tags --force
-        name: Force-fetch tags to work around actions/checkout#290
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -72,14 +72,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
```
abrown, line 74: Do you just want to move this into the new checkout action?

alexcrichton: I initially tried that yeah but ended up finding that using `./.github/...` requires checking out first so this was required to be split out (alas)

PR #478, opened 2024-08-21
Add llvm-dwp binary to toolchain

quanterion:
There is an option to compile debug information to separate file. It's supported in Clang via -gsplit-dwarf to generate dwo files but we need llvm-dwp tool to combine them into single dwp file supported by Chrome DWARF extension https://developer.chrome.com/blog/faster-wasm-debugging

It is already supported in Emscripten https://github.com/emscripten-core/emscripten/issues/13251

### DIFF
--- a/cmake/wasi-sdk-toolchain.cmake
+++ b/cmake/wasi-sdk-toolchain.cmake
@@ -45,6 +45,7 @@ set(tools
   llvm-ranlib
   llvm-strip
   llvm-dwarfdump
+  llvm-dwp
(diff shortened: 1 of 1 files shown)


yamt:
@sbc100 is https://github.com/WebAssembly/wasi-sdk/pull/364 still alive?

sbc100:
> @sbc100 is #364 still alive?

I'm not actively working on it so this change seems reasonable to land in the interim.

abrown (approved):

PR #482, opened 2024-08-23
test https://github.com/WebAssembly/wasi-libc/pull/526

yamt:


yamt:
in the windows job, for some reasons, the -isystem option in CFLAGS seems to prevent the linker from finding compiler-rt.
i don't understand how it happens because:
* afaik -isystem should not affect the linker behavior
* the compiler-rt path is explicitly given as BUILTINS_LIB

PR #487, opened 2024-09-05
Publish header with toolchain versions

abrown:
This change is an attempt to resolve [wasi-libc#490] but at the wasi-sdk level. It adds a `version.h` header file to
`share/wasi-sysroot/include/<target>/wasi` so that users have programmatic access to some extra information to output better error messages, e.g. This `version.h` looks something like:

```c
// Generated by wasi-sdk's `version.py` script.
#ifndef VERSION_H
#define VERSION_H

#define WASI_SDK_VERSION "24.6g754aec3d6f58+m"
#define WASI_LIBC_VERSION "b9ef79d7dbd4"

#endif
```

It _is_ a bit strange that we're adding to wasi-libc's include files from wasi-sdk (it would be cleaner if it happened in wasi-libc directly) but wasi-sdk actually has the information available.

[wasi-libc#490]: https://github.com/WebAssembly/wasi-libc/issues/490


cmake/wasi-sdk-sysroot.cmake
```diff
@@ -336,6 +336,19 @@ file(GENERATE OUTPUT ${version_file_tmp} CONTENT ${version_dump})
...
+  COMMAND ${PYTHON} ${version_script} header
+  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+  OUTPUT_VARIABLE version_header)
+foreach(target IN LISTS WASI_SDK_TARGETS)
+  message(STATUS "Creating version.h in ${version_dir}")
+  set(version_header_path ${wasi_sysroot}/include/${target}/wasi/version.h)
+  file(GENERATE OUTPUT ${version_header_path} CONTENT ${version_header})
+  add_custom_target(version-header-${target} DEPENDS ${version_header_path})
+  add_dependencies(build version-header-${target})
+endforeach()
```
sbc100, line 350: A more traditional cmake way of doing this would be to create a template file called `version.h` that looks like this:

```
#ifndef VERSION_H
#define WASI_SDK_VERSION @WASI_SDK_VERSION@
#define WASI_LIBC_VERSION @WASI_LIBC_VERSION@
#endif
```
And then use `configure_file` : https://cmake.org/cmake/help/latest/command/configure_file.html

Then you would just need to set `WASI_LIBC_VERSION` and `WASI_SDK_VERSION` as cmake variables (presumably still by running version.py.. but without needing to add anything new to version.py I would hope).

abrown: I like that better for the most part but I was worried about having a non-functional file cause problems. That approach means that we would add a non-functional `version.h` file to wasi-libc that kind of hangs around un-configured until it is seen by wasi-sdk. For those users who interact with wasi-libc directly (not through wasi-sdk), wouldn't they see an error if they included that un-configured `version.h`?

sbc100: Oh, sorry, the non-functional file is called `version.h.in`.. which is used to create the functional `version.h`, so they won't be confused.

sbc100:
> Thanks very much for looking into this! This would solve our problem very neatly!
> 
> To be clear, would I want to include `<version.h>` or would it be located elsewhere?

I think you would include `<wasi/version.h>`.   We certainly wouldn't want to call the file `version.h` if it was in the default include path, as that would be asking for trouble.

abrown:
> To be clear, would I want to include <version.h> or would it be located elsewhere?

@grynspan, I think it would be `wasi/version.h`?

yamt:
> It is a bit strange that we're adding to wasi-libc's include files from wasi-sdk (it would be cleaner if it happened in wasi-libc directly)

yes.
i'm against this because it introduces an incompatibility between wasi-sdk and direct use of wasi-libc.
(programs including this header won't build for the latter)

abrown:
If we did this exclusively in wasi-libc it could look like: https://github.com/WebAssembly/wasi-libc/pull/534.

grynspan:
Could the two be combined? Such that -libc provides its version and -sdk provides its version?

abrown:
I'm not really working on this any more.

PR #506, opened 2025-01-21
Update wasm-component-ld

alexcrichton:
Pulls in a fix for bytecodealliance/wasmtime#10058 in the adapters that are used by default.

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -7,7 +7,7 @@ runs:
     - name: Setup `wasmtime` for tests
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "18.0.2"
+        version: "29.0.1"
     - name: Install ccache, ninja (macOS)
       run: brew install ccache ninja
       if: runner.os == 'macOS'
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -7,7 +7,7 @@ runs:
     - name: Setup `wasmtime` for tests
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "22.0.1"
+        version: "29.0.1"
     - name: Install ccache, ninja (macOS)
       run: brew install ccache ninja
       if: runner.os == 'macOS'
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -7,7 +7,7 @@ runs:
     - name: Setup `wasmtime` for tests
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "29.0.1"
+        version: "22.0.1"
     - name: Install ccache, ninja (macOS)
       run: brew install ccache ninja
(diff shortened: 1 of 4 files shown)


alexcrichton:
I've updated the version of Wasmtime used in CI due to https://github.com/bytecodealliance/actions/issues/14 of what I suspect was causing the initial failure in CI for the first commit.

pchickey (approved):

alexcrichton:
The last failure was due to:

```
2025-01-22T01:42:30.6181255Z --- /dev/null	2025-01-21 23:59:07.888380651 +0000
2025-01-22T01:42:30.6181901Z +++ /src/build/sysroot/tests/general/wasm32-wasip1.-O0.no_arg_main.cc.wasm.stderr.observed	2025-01-22 01:42:30.599406609 +0000
2025-01-22T01:42:30.6182033Z @@ -0,0 +1 @@
2025-01-22T01:42:30.6182488Z +wasmtime: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by wasmtime)
2025-01-22T01:42:30.6182497Z 
2025-01-22T01:42:30.6183124Z 380/405 Test #380: test-wasm32-wasip1.-O2.no_arg_main.cc ...............................***Failed    0.01 sec
```

which means that the `wasmtime` binary was using a newer glibc than was found on the system. Wasmtime is built in an `almalinux:8` container with glibc 2.28 but the builds in this repository are done with `ubuntu:18.04` which has glibc 2.27. In the interest of not changing too much in one go I've tried updating Wasmtime from 18.0.2 (which doesn't work with the installation action) to 22.0.1 which the last version of Wasmtime built in an older container (`centos:7` which has since EOL'd)

I might send a second PR in the future to update the Linux containers to `almalinux:8` which I believe is a non-EOL'd distro if this works.

alexcrichton:
Hm ok now those failures are due to Wasmtime needing to be 23+ to run tests for components now (since WASI versions are now 0.2.3 instead of 0.2.0). I've pushed an update of the base image from Ubuntu 18.04 to 20.04 and I'll see what CI says...

alexcrichton:
Ok looks like CI is green  now.

@pchickey (or someone else) mind signing off on the update of the Linux base images from 18.04 to 20.04 for ubuntu? This will increase the glibc requirement which means that the artifacts will require slightly newer linux distros to run. I'm not sure how to otherwise easily work around this at this time.

pchickey:
I feel comfortable reviewing the wasm-component-ld bump but I haven't been in the trenches of linux/libc compat for a while to make that call, I'm going to tag in @abrown @sbc100 @sunfishcode (or others!) who have more experience in that front to sign off on that aspect

abrown:
IIRC, @sbc100 cared most about maintaining support for older images so that we knew the compatibility bounds.

sbc100:
20.04 focal seem reasonable to me.  We switch to that as our minimum a while back in emscripten.

alexcrichton:
Ok thanks all! I'm going to go ahead and merge this then 👍

PR #507, opened 2025-01-24
Pin CI runner images to explicit versions

alexcrichton:
CI is run infrequently enough on this repository that there's a nontrivial chance that GitHub updates the definition of `ubuntu-latest` for example between two CI runs. This can be confusing for contributors as PRs seemingly break CI when in reality they have nothing to do with the breakage and it's due to the image changing.

This PR pins all images to the definition of `*-latest` at this time. The one exception is that one builder was pinned to 22.04 and I'm going to update it to 24.04 here and see if I can't fix CI issues that come up.

This'll require explicit PRs to update these images in the future, but hopefully that's only once every few ~years so not too much of a burden.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -178,7 +178,8 @@ jobs:
             -DCMAKE_C_COMPILER=/usr/lib/llvm-18/bin/clang \
             -DCMAKE_SYSTEM_NAME=WASI \
             -DWASI_SDK_INCLUDE_TESTS=ON \
-            -DCMAKE_C_LINKER_DEPFILE_SUPPORTED=OFF
+            -DCMAKE_C_LINKER_DEPFILE_SUPPORTED=OFF \
+            -DCMAKE_CXX_LINKER_DEPFILE_SUPPORTED=OFF
       - run: ninja -C build
       - run: ctest --output-on-failure --parallel 10 --test-dir build/tests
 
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -165,25 +165,21 @@ jobs:
...
-      - run: cargo install wasm-component-ld@0.5.5
+      - run: cargo install wasm-component-ld@0.5.12
       - run: |
           cmake -G Ninja -B build -S . \
             -DCMAKE_C_COMPILER=/usr/lib/llvm-18/bin/clang \
             -DCMAKE_SYSTEM_NAME=WASI \
-            -DWASI_SDK_INCLUDE_TESTS=ON
+            -DWASI_SDK_INCLUDE_TESTS=ON \
+            -DCMAKE_C_LINKER_DEPFILE_SUPPORTED=OFF \
+            -DCMAKE_CXX_LINKER_DEPFILE_SUPPORTED=OFF
```
abrown, line 182: This is need here _and_ below?

abrown (approved):
Thanks!

.github/workflows/main.yml
alexcrichton: Unfortunately I believe so yeah. This layer is for the cmake build of wasi-sdk itself, aka when building tests. The cmake layer is then for the sub-cmake-builds of libraries like libcxx themselves. At least from CI it looked like both were necessary to get things passing when just using "clang" as the compiler.

PR #509, opened 2025-02-02
[ci] Use native arm runner instead of cross compiling

mcbarton:
With Ubuntu arm Github runners now available for general availability (see Github blogpost here https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ ) there is no longer a need to use the Ubuntu x86 runner and cross compile in the ci. This means you no longer need to skip building the sysroot on the ci for this platform.


mcbarton:
@alexcrichton can you review this PR, and if happy approve the ci to run? I'm not entirely sure I've named the PR well. Basically as far as I can tell from the ci your running a job which cross compiles for the arm architecture from the an x86 one (in this case Ubuntu). This feels unnecessary with Github recently making their Ubuntu arm runners generally available.

alexcrichton:
CI I think should be running now (there's a failure to debug). I think it's reasonable to do a native build, yeah. Although I also don't personally think it's much of a problem to do a cross-build either, it's nice to exercise the infrastructure to ensure it's working. Basically I have no preference either way myself.

.github/workflows/main.yml
```diff
@@ -22,23 +22,8 @@ jobs:
             os: ubuntu-24.04
 
           - artifact: arm64-linux
-            os: ubuntu-24.04
+            os: ubuntu-24.04-arm
```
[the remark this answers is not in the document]
alexcrichton: The actual builds happen in a docker container, so the installation line you're looking for is [here](https://github.com/WebAssembly/wasi-sdk/blob/62c3d46dd735e271f2f5f351c149ab1478dbe183/ci/docker/Dockerfile.arm64-linux#L4). That's one of the reasons this PR is failing on CI, the container/build scripts will need to be updated to run on arm instead of x64

ci/docker/Dockerfile.arm64-linux
```diff
@@ -1,11 +1,5 @@
 FROM wasi-sdk-builder-base
```
sbc100, line 1: Looks like we can probably just delete this file?

mcbarton: This is something I plan to do, alongside deleting x86 specific one. Plan to look at the ci this evening and work out why its failing.

sbc100 (approved):
lgtm if the tests pass!

ci/docker-build.sh
```diff
@@ -18,13 +17,7 @@ set -x
...
 # Build the base image which the image below can used.
 docker build \
-  --file ci/docker/Dockerfile.common \
-  --tag wasi-sdk-builder-base \
-  ci/docker
-
-# Build the container that is going to be used
-docker build \
-  --file ci/docker/Dockerfile.$1 \
+  --file ci/docker/Dockerfile \
```
sbc100, line 20: You can drop the `--file` now since this is the default Dockerfile.

ci/docker-build.sh
```diff
@@ -18,13 +17,7 @@ set -x
...
   --tag wasi-sdk-builder \
   ci/docker
```
sbc100, line 22: With `--file` removed I think this whole command can now be on single line?

Also, do we need the `--tag`?

ci/docker/Dockerfile
```diff
@@ -19,11 +19,12 @@ RUN apt-get update \
 
 # Install a more recent version of CMake than what 18.04 has since that's what
 # LLVM requires.
-RUN curl -sSLO https://github.com/Kitware/CMake/releases/download/v3.29.5/cmake-3.29.5-linux-x86_64.tar.gz \
-  && tar xf cmake-3.29.5-linux-x86_64.tar.gz \
-  && rm cmake-3.29.5-linux-x86_64.tar.gz \
+RUN export ARCH=$(uname -m) \
```
sbc100, line 22: No need for `export` here I think,

.github/workflows/main.yml
```diff
@@ -85,6 +69,13 @@ jobs:
           mkdir -p '${{ runner.tool_cache }}/ccache'
           echo 'CCACHE_DIR=${{ runner.tool_cache }}/ccache' >> $GITHUB_ENV
         shell: bash
+      # Remove and reinstall rust as default install seems broken on Ubuntu arm
+      - run: |
+            rm -rf $HOME/.rustup $HOME/.cargo
```
sbc100, line 74: No need for the `|` here if the command is just one line.  You can do `run: rm -rf $HOME/.rustup $HOME/.cargo` on the same line I think.

mcbarton:
Not sure what is happening with the ninja error at the moment. Trying reinstalling ninja to fix error like I did with rust without success. Will try and think of another solution.

mcbarton:
@sbc100 This PR is ready for another review. I tried removing the --file and --tags from the docker build/run commands, but hit an issue (Removed both together so don't know if I can run one without the other). It feels that there must be an elegant alternative to the if statement I added in the Dockerfile, but just couldn't seem to see it.

ci/docker-build.sh
```diff
@@ -16,17 +15,8 @@ fi
...
-  --tag wasi-sdk-builder-base \
-  ci/docker
-
-# Build the container that is going to be used
-docker build \
-  --file ci/docker/Dockerfile.$1 \
-  --tag wasi-sdk-builder \
-  ci/docker
+# Build the Docker imager
+docker build --file ci/docker/Dockerfile --tag wasi-sdk-builder ci/docker
```
sbc100, line 19: I'm pretty sure you can remove `--file ci/docker/Dockerfile`.. it is documented as being the default value.

.github/workflows/main.yml
```diff
@@ -89,6 +73,12 @@ jobs:
           mkdir -p '${{ runner.tool_cache }}/ccache'
           echo 'CCACHE_DIR=${{ runner.tool_cache }}/ccache' >> $GITHUB_ENV
         shell: bash
+      # Remove and reinstall rust as default install seems broken on Ubuntu arm
+      - run: rm -rf $HOME/.rustup $HOME/.cargo
+        if: matrix.os == 'ubuntu-24.04-arm'
+
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        if: matrix.os == 'ubuntu-24.04-arm'
```
sbc100, line 81: My only remaining question is why this is needed?  Where was the rust toolchain previously coming from before the use of `actions-rust-lang/setup-rust-toolchain` here?

mcbarton: I'm note entirely sure where in the repo the previous rust install was coming from, but any attempt by the ci at using it resulted in an error.

sbc100: It looks like maybe if you retore the `rust_target` then the rustup command below will run:

```
          if [ "${{ matrix.rust_target }}" != "" ]; then                             
            rustup target add ${{ matrix.rust_target }}                              
            cmake_args="$cmake_args -DRUST_TARGET=${{ matrix.rust_target }}"         
          fi    
```

Perhaps removing the rust_target was why this extra run was needed?

sbc100: If that doesn't work perhaps @alexcrichton has some clue?

mcbarton: I know this doesn't work, since its something I've already tried.

alexcrichton: Do you have a CI log for when the changes here weren't present? 

>  Where was the rust toolchain previously coming from before the use of actions-rust-lang/setup-rust-toolchain here?

GHA runners have `rustup`/`rustc`/`cargo` installed by default which is where the toolchain came from.

---

In the abstract though there are [known crashes](https://github.com/rust-lang/rust/issues/135867) in rustc on the new GHA runners for arm64. I'm not sure if that would affect this.

mcbarton: I reverted the change expecting to see the same error I saw previously, and the workflow ran without issue. Not sure what change I made since the error which fixed it.

sbc100 (approved):
Looks like it worked!  lgtm

.github/workflows/main.yml
saethlin: The crashes in question are random and seem to be an interaction between these CPUs and a kernel version that starts somewhere between Ubuntu 22 and Ubuntu 24. Use Ubuntu 22 if you want stable CI on these runners.

sbc100: Sounds reasonable, can we switch to Ubuntu 22 @mcbarton?

alexcrichton:
I'm not sure why, but why the latest commit downgrading from ubuntu 24 to 22? I agree with @sbc100 that prior to that this looks reasonable to me

saethlin:
@alexcrichton https://github.com/WebAssembly/wasi-sdk/pull/509#discussion_r1945488209

mcbarton:
Not sure whether I should revert back to Ubuntu 24.04 . Until I get confirmation either way i'll leave this PR as is with Ubuntu 22.04 arm.

sbc100:
Leaving the arm runner on 22.04 makes sense.   I think this change is good to land now, but I'll wait for @alexcrichton to confirm

alexcrichton:
Oh sorry I missed the reply from @saethlin in my emails, apologies!

PR #510, opened 2025-02-02
Update MacOS image ci to MacOS-15

mcbarton:
This PR adds to the ci jobs which cover MacOS 15. This allows the ci to cover the 2 most recent versions of MacOS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
                 -DRUST_TARGET=aarch64-unknown-linux-gnu
 
           - artifact: arm64-macos
-            os: macos-14
+            os: macos-15
             rust_target: aarch64-apple-darwin
             env:
               WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS: >-
@@ -49,7 +49,7 @@ jobs:
                 -DCMAKE_OSX_ARCHITECTURES=arm64
 
           - artifact: x86_64-macos
-            os: macos-14
+            os: macos-15
             rust_target: x86_64-apple-darwin
             env:
               WASI_SDK_CI_SKIP_SYSROOT: 1


sbc100 (commented):
Why not keep supporting 14?

mcbarton:
> Why not keep supporting 14?

Updated the PR to keep MacOS 14, and add MacOS 15 jobs alongside them, so that ci has support for both versions.

.github/workflows/main.yml
```diff
@@ -40,22 +40,39 @@ jobs:
...
                 -DCMAKE_OSX_ARCHITECTURES=x86_64
+          
+          - artifact: x86_64-macos-15
+            os: macos-15
+            rust_target: x86_64-apple-darwin
+            env:
+              WASI_SDK_CI_SKIP_SYSROOT: 1
+              WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS: >-
+                -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12
+                -DCMAKE_OSX_ARCHITECTURES=x86_64
```
sbc100, line 75: Is this needed though?   I would hope maybe we could just test on one OS version and rely on `CMAKE_OSX_DEPLOYMENT_TARGET=10.12` to ensure the result will run widely?

mcbarton: My experience with MacOS is that you can't take anything for granted (Although not specific to this repo, I have worked on ci in the past which worked for MacOS 14 but not MacOS 15). If the plan is to keep just one then I would keep the MacOS 15 runners, since I've found them just as reliable as the MacOS 14 ones while finding them faster.

alexcrichton:
Personally I think it's reasonable to update. GitHub Actions images will continue to update over time and additionally phase out older images, so AFAIK it's inevitable that updates are needed. When the updates happen shouldn't impact binary releases too much due to `-DCMAKE_OSX_DEPLOYMENT_TARGET=..`. 

To me there's a dual risk either way of if we don't test the latest image then there's a risk it doesn't work. Similarly if we stop testing the old image there's a risk it stops working. Given the CI resources for this repository though I'd say we should only test one and the risks involved here are small enough that it probably doesn't matter which is tested.

One final note is that by changing the names of the jobs it'll require reconfiguring this repository to expect a different job name with "required" checks. If possible can the job names stay the same to avoid needing to reconfigure?

.github/workflows/main.yml
```diff
@@ -40,16 +40,16 @@ jobs:
...
+            os: macos-15
             rust_target: aarch64-apple-darwin
             env:
               WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS: >-
                 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.12
                 -DCMAKE_OSX_ARCHITECTURES=arm64
-
-          - artifact: x86_64-macos
-            os: macos-14
+
```
sbc100, line 50: Looks like some accidental whitespace added here?

sbc100 (approved):

mcbarton:
Closing as it was chosen to keep older image in order to keep compatibility with older operating systems.

PR #511, opened 2025-02-02
Update Windows image CI to Windows-2025

mcbarton:
This PR adds a Windows 2025 job to the ci, so that the 2 most recent versions of Windows server which are available as Github runners are supported by the ci.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,7 +58,7 @@ jobs:
                 -DCMAKE_OSX_ARCHITECTURES=x86_64
 
           - artifact: x86_64-windows
-            os: windows-2022
+            os: windows-2025
 
     env: ${{ matrix.env || fromJSON('{}') }}
     steps:


mcbarton:
@sbc100 pinging for a review.

sbc100 (commented):
Is it important to test on two different version of windows?  Presumably there is cost to each config we add/test.

mcbarton:
> Is it important to test on two different version of windows? Presumably there is cost to each config we add/test.

I did it to match what I change the MacOS PR to https://github.com/WebAssembly/wasi-sdk/pull/510 . If you mean in terms of monetary cost, no. For open source projects I've found you generally get unlimited minutes for Github runners for free. For the open source projects I maintain Github we use the runners labelled as generally available, and have never hit a limit imposed by Github.

sbc100:
> > Is it important to test on two different version of windows? Presumably there is cost to each config we add/test.
> 
> I did it to match what I change the MacOS PR to #510 . If you mean in terms of monetary cost, no. For open source projects I've found you generally get unlimited minutes for Github runners for free. For the open source projects I maintain Github we use the runners labelled as generally available, and have never hit a limit imposed by Github.

I think the limits can show up as throttling, and a general slowdown the CI throughput, but I'm not totally sure how github manage it.  I also think maybe they do some kind of per-org metering too.  We've been seeing WebAssembly/binaryen CI going super slow recently.  I think as a general rule the more work you do in CI the slower you can expect your CI cycle to be (I think github have some kind of metering that makes this true even you do all of your work in parallel on many runners).

mcbarton:
> > > Is it important to test on two different version of windows? Presumably there is cost to each config we add/test.
> > 
> > 
> > I did it to match what I change the MacOS PR to #510 . If you mean in terms of monetary cost, no. For open source projects I've found you generally get unlimited minutes for Github runners for free. For the open source projects I maintain Github we use the runners labelled as generally available, and have never hit a limit imposed by Github.
> 
> I think the limits can show up as throttling, and a general slowdown the CI throughput, but I'm not totally sure how github manage it. I also think maybe they do some kind of per-org metering too. We've been seeing WebAssembly/binaryen CI going super slow recently. I think as a general rule the more work you do in CI the slower you can expect your CI cycle to be (I think github have some kind of metering that makes this true even you do all of your work in parallel on many runners).

You can get a slow down as an organisation if your hitting the limit of the maximum number of concurrent Github runners limit allowed (outlined here https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration ). These MacOS limit is generally very easy to hit, and you can be waiting a while for MacOS runner if many PRs in an organisation are simultaneously trying to use them. My suggestion is to change my Windows and MacOS PRs back to using the latest versions (which I have found to be slightly quicker). That means as a rule you'll get through each target system job quicker, using less minutes, and less runners overall.

alexcrichton:
Personally I would agree with @sbc100 that we probably don't need both tested here (or for macos). It eats up comcurrency in the org for quite awhile without too too much benefit. I think it's reasonable to update the image though.

mcbarton:
> Personally I would agree with @sbc100 that we probably don't need both tested here (or for macos). It eats up comcurrency in the org for quite awhile without too too much benefit. I think it's reasonable to update the image though.

Sometime later today I will revert my PRs back to using just updating to the latest image.

sbc100 (changes requested):
Actually I wonder if its better to keep testing with older version of windows to make sure we are compatible?

Do you have some reason to prefer/require this very latest version?  I don't think wasi-libc should be depending it, right?

mcbarton:
@alexcrichton @sbc100 I've changed this PR and the MacOS PR to just change the image to the latest version, but not the name. That is the consenous I think has been reached. Let me know if it was decided to keep the older images for compatibility checks.

sbc100:
Can you respond to https://github.com/WebAssembly/wasi-sdk/pull/511#pullrequestreview-2590271633  ?  Basically, I don't think we should be doing this unless there is good reason.

mcbarton:
@sbc100 If you look at the Windows job that just run on main the ci took over 2 hours to run on the Windows-2022 runner (see https://github.com/WebAssembly/wasi-sdk/actions/runs/13116480943/job/36591949707#step:12:1). If you look at the same job on the Windows-2025 runner here https://github.com/WebAssembly/wasi-sdk/actions/runs/13113946448/job/36594227539#step:12:1 it took around 40 minutes. That speed up seems worth it if your trying to save runner minutes, and effect the slowdown of other repos in the organisation.

sbc100:
Thats interesting.  Are you sure the differences is accounted for by the version of windows being used?  It seems odd that 2025 would be so much faster.  Do you know if that is expected?   If so, please update the PR description explaining that 2025 is much faster (and if possible why its faster).

mcbarton:
> Thats interesting. Are you sure the differences is accounted for by the version of windows being used? It seems odd that 2025 would be so much faster. Do you know if that is expected? If so, please update the PR description explaining that 2025 is much faster (and if possible why its faster).

I have no idea why its so much faster to be honest. I have had this in the past with other operating systems. This is normally due to an increase in the number of cpu cores available to the new runner, but cannot find any documentation on the number of cores of the separate runners, so on my part its complete speculation. I'll rebase the PR incase the speed difference is due to a recent change to main.

sbc100:
Lets close this then.  Better to continue to support reasonably old windows unless we have reason not to.

mcbarton:
> Lets close this then. Better to continue to support reasonably old windows unless we have reason not to.

Does the same apply to my MacOS PR?

sbc100:
> > Lets close this then. Better to continue to support reasonably old windows unless we have reason not to.
> 
> Does the same apply to my MacOS PR?

I think so yes, unless you have a specific reason that we shouldn't keep testing on older OS versions.

mcbarton:
Closing as it was chosen to keep older image in order to keep compatibility with older operating systems.

PR #513, opened 2025-02-03
[ci] Cancel outstanding workflows when new changes are added to PRs 

mcbarton:
If you submit a commit to a PR at the moment and then find an issue with that commit, and submit a new commit to fix the PR before the ci has finished then the ci runs fully for both commits. This PR makes sure that if a new commit is submitted the old ci run is cancelled. This should save on Github runner minutes, ensuring that the slow down across the organisation due to Github runner consumed minutes is less pronounced.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,6 +10,10 @@ on:
   pull_request:
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: Build ${{ matrix.artifact }}


mcbarton:
@sbc100 pinging for review. Its the same change I made in wasi-libc here https://github.com/WebAssembly/wasi-libc/pull/571

sbc100:
The folks over in the binaryen repo are saying that they have observed this to be the default: https://github.com/WebAssembly/binaryen/pull/7269.   Are you sure its not happening already today?

mcbarton:
@sbc100 This is not the default. Here is an example. In this PR https://github.com/WebAssembly/wasi-sdk/pull/509 I put in a commit 1 hour ago https://github.com/WebAssembly/wasi-sdk/actions/runs/13137913398 and it was broken, so I put in another commit shortly afterwards before this ci run had finished here https://github.com/WebAssembly/wasi-sdk/actions/runs/13138053034 . As you can see, the first ci run carried on running despite me putting in a new commit. Also the default behaviour is explained here in the Github documentation https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior , which is not to cancel in progress jobs.

PR #517, opened 2025-02-22
[ci] Update submodules independently ci

mcbarton:
I'm not so sure how this PR will be received (since it increases the number of lines in the action), but decided to put in anyway. In the checkout/action.yml file there is a comment that --depth 1 cannot be used since the config submodule is not fixed to a tag/branch. Given the size of llvm-project its probably worth separating out the config submodule update, so that other submodules can be updated with depth 1. This may lead to a small increase (not sure how noticeable until ci runs) in speed of the ci, given the size of the llvm-project submodule. I also admit I may be completely wrong in this assumption.

### DIFF
--- a/.github/actions/checkout/action.yml
+++ b/.github/actions/checkout/action.yml
@@ -7,10 +7,14 @@ runs:
     - run: git fetch --tags --force
       name: Force-fetch tags to work around actions/checkout#290
       shell: bash
-    # We can't use `--depth 1` here sadly because the GNU config
-    # submodule is not pinned to a particular tag/branch. Please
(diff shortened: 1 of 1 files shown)


sbc100:
But doesn't `--depth=64` just mean that the most recent 64 commits a pulled?   

I can't imagine the pulling 64 commits out of llvm's 528312 total commits, is any more expensive than just pulling 1.   Are you seeing a difference?    I imagine that amount of data saved here would be just a few bytes out of several gb.

TerrorJack:
Or maybe just bump `config` if needed?

PR #523, opened 2025-04-16
SetjmpLongjmp.md: update for LLVM 20

yamt:


yamt:
> In talking to @alexcrichton, @sbc100, @sunfishcode: this seems like a reasonable change given the state of LLVM 20. We were confused by the `phase3` nomenclature; can we swap that out with `legacy`? It would also be good to add some more context to warn users about relying on this, since LLVM 21 may make this obsolete.

thank you. i added some text to make it clearer.

abrown (approved):

Issue #522, opened 2025-03-22
windows arm64 release


JakeBeaver:
I cannot build a dotnet 8.0 application with a `RuntimeIdentifier` of `wasi-wasm` on my arm64 windows machine, because that process requires the following file to exist: 

`https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-24/wasi-sdk-24.0-arm64-windows.tar.gz`

Can this version also be built and packaged with releases?

sbc100:
Current we don't build or test wasi-sdk on windows arm64.   How hard would it be to add support for windows arm64 to our github actions?  I really don't know.   If you would like to look into adding this I'm sure PRs would be most welcome.

mcbarton:
Windows arm Github runners for open source projects are on their roadmap (see https://github.com/orgs/github/projects/4247?pane=issue&itemId=94973781&issue=github%7Croadmap%7C1098). Once they become available I may look into this issue.

PR #524, opened 2025-04-17
Add Windows 11 arm support to ci

mcbarton:
This PR should hopefully fix https://github.com/WebAssembly/wasi-sdk/issues/522 . According to https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/ Windows arm Github runners are now available for free for public repositories. Therefore I have added them to wasi-sdks ci.

### DIFF
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -16,6 +16,8 @@ runs:
       run: choco install ccache ninja
       if: startsWith(matrix.os, 'windows')
       shell: bash
+      # Windows arm runners don't come with rust by default (see https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md)
+      # but the x86 ones do (see https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md)
     - name: Install cargo (Windows-arm)
       run: choco install rust
       if: matrix.os == 'windows-11-arm'
--- a/.github/actions/install-deps/action.yml
+++ b/.github/actions/install-deps/action.yml
@@ -14,7 +14,11 @@ runs:
       shell: bash
     - name: Install ccache, ninja (Windows)
       run: choco install ccache ninja
-      if: runner.os == 'Windows'
+      if: startsWith(matrix.os, 'windows')
+      shell: bash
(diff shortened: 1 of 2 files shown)


mcbarton:
Will investigate the failures here https://github.com/WebAssembly/binaryen/pull/7517/checks#step:12:90 and here https://github.com/WebAssembly/binaryen/pull/7517/checks#step:12:92 tomorrow if I can

mcbarton:
@sbc100 can you manually rerun the failing workflow jobs? The 2 failing jobs are due an issue with the runners internet connection when the job started (see error in output messages in workflow runs). Rerunning the failed jobs they should pass, and then this PR is ready for review.

alexcrichton:
I hit retry

mcbarton:
@alexcrichton @sbc100 this PR is ready for review.

sbc100 (approved):

.github/actions/install-deps/action.yml
```diff
@@ -14,7 +14,11 @@ runs:
       shell: bash
     - name: Install ccache, ninja (Windows)
       run: choco install ccache ninja
-      if: runner.os == 'Windows'
+      if: startsWith(matrix.os, 'windows')
+      shell: bash
+    - name: Install cargo (Windows-arm)
+      run: choco install rust
+      if: matrix.os == 'windows-11-arm'
```
sbc100, line 21: Why is this needed on arm-windows, only?  Maybe worth a comment?

mcbarton: If I remember correctly its because the Windows arm runners don't come with rust by default (see https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md) , but the x86 ones do (see here https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md ). Would this comment in the action.yml file suffice?

sbc100: Yes a comment right here seems like a good idea.

mcbarton: Done.

yamt (approved):

mcbarton:
Is it possible for this to go in now?

PR #533, opened 2025-06-11
Enable per-target runtime directory for compiler-rt

kateinoigakukun:
This change allows us to have per-target libclang_rt.* libraries instead of per-OS. This is particularly required for w/threads and wo/threads targets, whose rt libraries could be incompatible even with the same OS string.

This change starts building `compiler-rt` for w/threads and wo/threads separately, so that we can enable thread-related features for p1-threads target for building upcoming ASan clang_rt libraries. Also `COMPILER_RT_OS_DIR` is not set anymore as it's unused when `LLVM_ENABLE_PER_TARGET_RUNTIME_DIR` is enabled.

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -66,31 +66,38 @@ endif()
 # compiler-rt build logic
 # =============================================================================
 
-ExternalProject_Add(compiler-rt-build
-  SOURCE_DIR "${llvm_proj_dir}/compiler-rt"
-  CMAKE_ARGS
-      ${default_cmake_args}
-      -DCOMPILER_RT_BAREMETAL_BUILD=ON
-      -DCOMPILER_RT_BUILD_XRAY=OFF
-      -DCOMPILER_RT_INCLUDE_TESTS=OFF
-      -DCOMPILER_RT_HAS_FPIC_FLAG=OFF
-      -DCOMPILER_RT_DEFAULT_TARGET_ONLY=ON
-      -DCOMPILER_RT_BUILD_SANITIZERS=OFF
-      -DCOMPILER_RT_BUILD_XRAY=OFF
-      -DCOMPILER_RT_BUILD_LIBFUZZER=OFF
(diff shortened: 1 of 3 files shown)


alexcrichton:
I think [this may need updating too](https://github.com/WebAssembly/wasi-sdk/blob/38c97448e00facd18826bc73b5bb6c69219ffb4d/ci/merge-artifacts.sh#L65-L70) but otherwise seems resonable to me 👍, thanks!

sbc100 (approved):

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -360,7 +370,7 @@ set(dist_dir ${CMAKE_CURRENT_BINARY_DIR}/dist)
 # Tarball with just `compiler-rt` builtins within it
 wasi_sdk_add_tarball(dist-compiler-rt
   ${dist_dir}/libclang_rt.builtins-wasm32-wasi-${wasi_sdk_version}.tar.gz
-  ${wasi_resource_dir}/lib/wasi)
+  ${wasi_resource_dir}/lib/wasm32-unknown-wasi)
```
sbc100, line 373: Is it normall for the target to appear both in the directory name `${wasi_resource_dir}/lib/wasm32-unknown-wasi` *and* in the library name `libclang_rt.builtins-wasm32-wasi`?  It seems somewhat redundant.

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -106,12 +113,15 @@ add_custom_target(compiler-rt-post-build
   COMMAND ${CMAKE_COMMAND} -E copy_directory
     ${clang_resource_dir}/include ${wasi_resource_dir}/include
 
-  # Copy the `lib/wasi` folder to `libc/wasi{p1,p2}` to ensure that those
+  # Copy the `lib/wasm32-unknown-wasi` folder to `libc/wasm32-unknown-wasi{p1,p2}` to ensure that those
```
sbc100, line 116: Should `libc` here be just `lib`?  Maybe a pre-existing typo?

PR #544, opened 2025-07-08
Update .gitmodules

FaFre:
Switch to the more stable git server, since the current one often does have some reliability issues.


FaFre:
Hi Alex, while some issues regarding availability of savannah git recommend to use the https mirror, I couldn't find any recommendations from official GNU side that support the use of this mirror. Opening this PR was a bit too hasty, and I guess the old one is actually the widest used one. So I will be closing this PR.

PR #564, opened 2025-09-15
Always build wasm32-wasip2 with `-fPIC`

alexcrichton:
This commit updates the build of the `wasm32-wasip2` target to always pass `-fPIC` to compilation of wasi-libc. This notably enables using `libc.a` in dynamic linking situations instead of being forced to use `libc.so`. While many applications likely want to use `libc.so` I've found it more flexible if objects are compatible with as many builds as possible. This will enable, for example, building shared libraries in Rust by default since Rust only ships `libc.a`, not `libc.so`, in the pre-built sysroot. This behavior additionally matches the Rust `wasm32-wasip2` target where `-fPIC` is enabled by default there.


src/wasi-libc
alexcrichton, line 1: This updates wasi-libc to pull in https://github.com/WebAssembly/wasi-libc/pull/624

cmake/wasi-sdk-sysroot.cmake
alexcrichton, line 1: I went ahead and refactored this a bit to reduce the duplication between the various branches to also pass `-fPIC` unconditionally on wasip2

alexcrichton:
Blocked on https://github.com/WebAssembly/wasi-libc/issues/625 for tests

dicej (approved):

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -139,31 +139,32 @@ add_custom_target(compiler-rt DEPENDS compiler-rt-build compiler-rt-post-build)
...
+  elseif(NOT ${target} MATCHES threads)
+    list(APPEND extra_make_flags libc_so)
+  endif()
+
+  if(${target} MATCHES threads)
+    list(APPEND extra_make_flags THREAD_MODEL=posix)
+  elseif(${target} MATCHES p2)
+    list(APPEND extra_make_flags WASI_SNAPSHOT=p2)
+    # Always enable `-fPIC` for the `wasm32-wasip2` target. This makes `libc.a`
+    # more flexible and usable in dynamic linking situations.
```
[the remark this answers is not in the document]
alexcrichton: Can you detail why that's a better approach? That still suffers from "align the correct matrix of flags and objects" I'm trying to avoid.

PR #566, opened 2025-10-03
Remove `wasm32-wasi` and `wasm32-wasi-threads` targets

alexcrichton:
These are replaced by `wasm32-wasip1` and `wasm32-wasip1-threads`, respectively. The intention is to free up the namespace of "wasm32-wasi" for a hypothetical future target of WASI 1.0. This naming matches the Rust naming scheme which dropped support for `wasm32-wasi` some time ago (and renamed it to `wasm32-wasip1`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -259,6 +259,6 @@ jobs:
           cmake -G Ninja -B build -S . \
             -DWASI_SDK_INCLUDE_TESTS=ON \
             -DWASI_SDK_TEST_HOST_TOOLCHAIN=ON \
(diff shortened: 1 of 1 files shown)


TerrorJack:
This change is too early and too disruptive for downstream wasi-sdk users, many scripts still use `wasm32-wasi-clang` etc. It's better to revisit this after wasi 1.0 really comes out and picks up community momentum.

pchickey:
[This change was made in Rust in April](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets/), and downstream users adjusted. The sky did not fall.

The best time to make this change was yesterday, and the second-best is now. Wasi-sdk adoption is presently at the lowest adoption it will be between now and when wasi 1.0 comes out.

PR #580, opened 2025-12-19
Update wasi-libc to cmake-enabled build

alexcrichton:
Updating to handle https://github.com/WebAssembly/wasi-libc/commit/688a685d1c5037ef8d718b9d3f91a4063c3ace77 from wasi-libc


pchickey (approved):
🎉

Issue #565, opened 2025-10-02
Support for C++ exceptions


alexcrichton:
I noticed that there wasn't already a tracking issue for C++ exceptions, so I wanted to open one up to at least track the progress of this. Currently wasi-sdk does not support C++ exceptions, and trying to fiddle with the build locally shows that it's not a turnkey solution at this time or a small PR to some build configuration. Some questions that I ran into locally are:

* Should exceptions support be provided in existing targets, or a new target? My assumption is that the entire C++ standard library needs to be built with/without exceptions meaning that technically `-fexceptions` wants an entirely different sysroot than `-fno-exceptions`. I don't believe that Clang currently dynamically selects a different sysroot based on this flag, so the next best thing we could do would be to add more targets. That would double the size of the target list, however.
* Right now using the standard wasm exceptions proposal requires `-mllvm -wasm-use-legacy-eh=false`, and I'm not sure if that would require tripling the sysroot where one uses legacy eh, one doesn't, and one uses no eh.
* Currently `-fwasm-exceptions` looks to be required, but I'm not sure if this is a temporary state of affairs or one that's intended to last.
* Wasi-sdk looks like it will need to build LLVM's libunwind. There's a few build issues with this where it doesn't natively build for wasm as-is (even with all exception-related things turned on). It's not clear to me if (a) libunwind is not expected to build for wasm, (b) libunwind should be buildable as-is but maybe with different configuration, or (c) perhaps there are Emscripten patches to libunwind which haven't made their way into LLVM itself.
* I don't know where the responsibility of passing `-lunwind` lies. Should the Clang driver automatically pass that? Should users be required to pass that?

These are the questions I've run into locally myself. From an implementation perspective all the "meat" of C++ exceptions is done in the sense that I was able to edit enough files to get something compiling and executing. The remaining bits of work are all on the toolchain side I believe (all of the above questions)

whitequark:
See also https://github.com/WebAssembly/wasi-sdk/issues/334, which is less detailed than this issue but was filed earlier.

yerzham:
Ok, my attempts to fix the older version of llvm project were irrelevant, sorry. As a note for people like me who used the latest release of wasi-sdk (28) and wanted to understand if and when C++ exceptions are going to be supported: 

[...]

PR #590, opened 2026-02-07
Add experimental support for C++ exceptions

alexcrichton:
> **Note**: this PR is me weaving together the work of many others. For example @yerzham's and @cpetig's work on #565 pretty much shaped this PR. Thank you!

This commit adds initial configuration support for building wasi-sdk with support for C++ exceptions in WebAssembly. A small test is included which is exercised in CI at this time, but otherwise this does not update CI to actually ship sdk builds with C++ exceptions enabled. Instead the intention here is to make it easier to test builds with support for C++ exceptions and centralize planning/development around this.

The goal here is to get things compiling to the point that applications can be compiled. I haven't thoroughly tested C++ exceptions and as evident in this PR it still requires changes in LLVM. Some small logic is added here to apply a `*.patch` file to LLVM to avoid needing a new submodule fork or waiting for upstream changes. The intention is that this `*.patch` is short-lived once the changes are officially merged into LLVM itself.

The `*.patch` included here contains llvm/llvm-project#168449 as well as another minor edit I found was necessary to get things compiling locally. Given the discussion on that PR it looks like once LLVM is updated with that merged the extra part of the `*.patch` won't be necessary.

This PR notably does not enable shared libraries when exceptions are enabled. I don't know enough about how things are supposed to work to be able to fully diagnose the compilation errors I'm seeing if shared libraries are enabled. This is something I'd hope would be fixed before actually shipping exceptions support.

This PR then additionally folds in [this gist][gist] for various bits of build logic to this repository itself.

Finally, this PR includes some documentation about the current status of exceptions and links to various tracking issues too.

[gist]: https://gist.github.com/yerzham/302efcec6a2e82c1e8de4aed576ea29d

### DIFF
--- a/tests/general/exceptions.cc
+++ b/tests/general/exceptions.cc
@@ -1,12 +1,17 @@
 #include <iostream>
+#include <stdlib.h>
 
 int main() {
 #ifdef __wasm_exception_handling__
   try {
     throw std::runtime_error("An error occurred");
+    abort();
   } catch (const std::runtime_error& e) {
     // ..
+    return 0;
   }
-#endif
+  abort();
+#else
   return 0;
+#endif
 }
--- a/CppExceptions.md
+++ b/CppExceptions.md
@@ -35,7 +35,8 @@ Without these flags programs using C++ exceptions will not work correctly:
 This can be specified for example with:
 
 ```shell script
-$ export CFLAGS="-fwasm-exceptions -mllvm -wasm-use-legacy-eh=false -lunwind"
+$ export CFLAGS="-fwasm-exceptions -mllvm -wasm-use-legacy-eh=false"
+$ export LDFLAGS="-lunwind"
 ```
 
 ## Limitations


tests/general/exceptions.cc
```diff
@@ -0,0 +1,12 @@
+#include <iostream>
+
+int main() {
+#ifdef __wasm_exception_handling__
+  try {
+    throw std::runtime_error("An error occurred");
+  } catch (const std::runtime_error& e) {
+    // ..
```
ricochet: `std::cout << "caught" << std::endl;` with a `.stdout.expected` to validate the exception  exception round-tripped correctly

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -290,6 +310,9 @@ function(define_libcxx_sub target target_suffix extra_target_flags extra_libdir_
     USES_TERMINAL_CONFIGURE ON
     USES_TERMINAL_BUILD ON
     USES_TERMINAL_INSTALL ON
+    PATCH_COMMAND
```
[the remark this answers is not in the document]
cpetig: As this fixes a clear mistake (use windows dll decoration by default, thus also on wasm) the fix could be unconditional. The assembly.h part will become unnecessary (although still correct) in future llvm versions.

tests/general/exceptions.cc
cpetig: Second thought, perhaps place the `return 0;` inside the catch clause and `return 1;` in the fall-through path?

whitequark:
Would it be possible to upload exception-ful WASI SDK as a perhaps unstable tag? I would like to start using it in YoWASP ASAP, but the overhead of having to build my own copy of the SDK first means I can't really justify it at the moment. It doesn't need to be official or supported, just built and available for a download.

PR #606, opened 2026-03-11
Build a sysroot that supports C++ exceptions by default

### LINKED ISSUE #334: Add Support for WASM Exceptions
There are (plans to add support for exceptions in WAMR)[https://github.com/bytecodealliance/wasm-micro-runtime/issues/1884]. Are there any plans to add support to the WASI-SDK tool chain to support Native WASM Exceptions?

### LINKED ISSUE #565: Support for C++ exceptions

alexcrichton:
This commit collects together some LLVM PRs, some changes in the build configuration here, and some thoughts from #565 and related issues. Specifically the changes here are:

* The patch for llvm/llvm-project#168449 is updated to its upstream (unlanded) form.
* Patches for the (landed) llvm/llvm-project#185770 and llvm/llvm-project#185775 are added.
* The `WASI_SDK_EXCEPTIONS` configuration is now either `ON`, `OFF`, or `DUAL`. The default depends on the version of Clang in use, where 23.0.0+ (which isn't released officially yet) will be `DUAL` and otherwise it's `OFF`. CI for our custom-built patched toolchain defaults to `DUAL`.
* In `DUAL` mode libcxx is built twice into two different directories, once with exceptions and once without. This is supported by LLVM patches and means that Clang will select the right set of libraries based on compiler flags.

The end result here is that the produced toolchain from this repository, by default, supports C++ exceptions. Additionally if exceptions-related flags are not passed then the final binary will not use C++ exceptions nor require the wasm exception-handling proposal.

There's still follow-up work from #565, such as:

* Subjectively it feels wordy to pass `-fwasm-exceptions` vs `-fexceptions`.
* Personally I think `-mllvm -wasm-use-legacy-eh=false` should become the default upstream.
* Subjectively I don't think that `-lunwind` should be necessary and it should be injected automatically with `-fwasm-exceptions` (or `-fexceptions`).
* Shared libraries for exceptions remain disabled due to build errors I do not personally know how to resolve.

I'll file follow-up issues for these once this has landed since they're more minor compared to the main body of "anything works".

Closes #334
Closes #565

### DIFF
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -381,7 +381,7 @@ function(define_libcxx target)
   # compiled and placed into the sysroot. They're named slightly differently to
   # have unique CMake rules.
   #
-  # Otherwise there's only one build of libcxx and it's either got exceptions for
+  # Otherwise there's only one build of libcxx and it's either got exceptions or
   # it doesn't depending on configuration.
   if (WASI_SDK_EXCEPTIONS STREQUAL "DUAL")
     define_libcxx_and_lto(${target} "" OFF)
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -377,6 +377,12 @@ endfunction()
 function(define_libcxx target)
   add_custom_target(libcxx-${target})
 
+  # For dual-mode exceptions-and-not there are two versions of libcxx which are
+  # compiled and placed into the sysroot. They're named slightly differently to
+  # have unique CMake rules.
+  #
+  # Otherwise there's only one build of libcxx and it's either got exceptions for
+  # it doesn't depending on configuration.
   if (WASI_SDK_EXCEPTIONS STREQUAL "DUAL")
     define_libcxx_and_lto(${target} "" OFF)
     define_libcxx_and_lto(${target} "-exn" ON)
--- a/cmake/wasi-sdk-sysroot.cmake
+++ b/cmake/wasi-sdk-sysroot.cmake
@@ -20,12 +20,27 @@ message(STATUS "Found executable for `ar`: ${CMAKE_AR}")
 
 find_program(MAKE make REQUIRED)
 
+set(EXCEPTIONS_DEFAULT "OFF")
+if(CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 23.0.0)
+  set(EXCEPTIONS_DEFAULT "DUAL")
+endif()
+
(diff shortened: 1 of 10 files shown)


ricochet (approved):
A few non-blocking nits.

cmake/wasi-sdk-sysroot.cmake
```diff
@@ -327,30 +351,53 @@ function(define_libcxx_sub target target_suffix extra_target_flags extra_libdir_
...
+endfunction()
+
+function(define_libcxx target)
+  add_custom_target(libcxx-${target})
+
+  # For dual-mode exceptions-and-not there are two versions of libcxx which are
+  # compiled and placed into the sysroot. They're named slightly differently to
+  # have unique CMake rules.
+  #
+  # Otherwise there's only one build of libcxx and it's either got exceptions for
```
dicej:
```suggestion
  # Otherwise there's only one build of libcxx and it's either got exceptions or
```

dicej (approved):
Thanks!

Issue #607, opened 2026-03-11
Add riscv64 host platform support


gounthar:
## Summary

WASI SDK currently ships pre-built binaries for six host platforms (x86_64 and arm64 on Linux, macOS, Windows). I'd like to request adding `riscv64-linux` as a supported host platform.

## Context

I successfully built WASI SDK v30 from source on native RISC-V hardware (Banana Pi F3, SpacemiT K1, rv64gc, 16 GB RAM, Debian Trixie). The build completes with **zero source code patches** ; only cmake flags and one post-install symlink are needed.

The full process is documented in a blog post: [Building WASI SDK on RISC-V](https://bruno.verachten.fr/2026/02/25/Building-WASI-SDK-on-RISC-V/)

## Four workarounds required

All are cmake flags or symlinks, no source modifications:

1. **Stage 2 needs Stage 1 Clang** (not system GCC):
   `-DCMAKE_C_COMPILER="$INSTALL_DIR/bin/clang"` / `-DCMAKE_CXX_COMPILER="$INSTALL_DIR/bin/clang++"`

2. **Skip native compiler test** (Stage 1 Clang only targets wasm32):
   `-DCMAKE_C_COMPILER_WORKS=ON` / `-DCMAKE_CXX_COMPILER_WORKS=ON`

3. **Skip WASIP2 targets** (wit-bindgen/wasm-tools/wkg don't ship riscv64 binaries):
   `-DWASI_SDK_TARGETS="wasm32-wasi;wasm32-wasip1"`

4. **Fix resource dir path**: symlink `clang-resource-dir/lib` → `lib/clang/21/lib`

## Build stats

| Metric | Value |
|--------|-------|
| Stage 1 (LLVM+Clang+LLD) | ~8-10 hours |
| Stage 2 (wasi-libc sysroot) | ~15-20 minutes |
| Installed SDK size | 244 MB |
| hello-world.wasm output | 105 KB |
| Source patches needed | **0** |
| WASI SDK version built | 30.2g1033443e5c36 |
| Clang version | 21.1.4-wasi-sdk |

## What this enables

With WASI SDK on riscv64, developers can do fully self-hosted WebAssembly development on RISC-V: C/C++ source → Clang (native rv64) → .wasm → iwasm (native rv64), with no x86 or ARM host needed.

This is part of a broader effort proving the full WebAssembly stack works on RISC-V, including WAMR and the Ocre runtime.

## Proposal

I'd be happy to contribute:
[...]

PR #608, opened 2026-03-11
docs: add riscv64 native build instructions

gounthar:
## Summary

Add documentation for building WASI SDK from source on native riscv64 Linux hardware.

The build completes with **zero source patches** — only cmake configuration flags are needed:

1. `-DCMAKE_C_COMPILER_WORKS=ON` to skip compiler test (Stage 1 Clang only targets wasm32)
2. `-DWASI_SDK_TARGETS="wasm32-wasi;wasm32-wasip1"` to skip WASIP2 (wit-bindgen/wasm-tools/wkg don't ship riscv64 binaries)
3. Post-install symlink for the Clang resource directory

Tested on Banana Pi F3 (SpacemiT K1, rv64gc, 8 cores, 16 GB RAM, Debian Trixie) with WASI SDK v30. Full self-hosted loop verified: C → Clang (native rv64) → .wasm → iwasm (native rv64).

Ref: #607

## What this enables

Self-hosted WebAssembly development on RISC-V hardware without needing an x86 or ARM host for compilation.

## Test plan

- [x] Built WASI SDK v30 on Banana Pi F3 (rv64gc) following these instructions
- [x] Compiled hello-world.c to .wasm using the native riscv64 toolchain
- [x] Ran the .wasm with iwasm on the same board

### DIFF
--- a/docs/riscv64-build.md
+++ b/docs/riscv64-build.md
@@ -1,0 +1,115 @@
+# Building WASI SDK on riscv64
+
+WASI SDK can be built from source on native riscv64 Linux hardware. The build
+completes with zero source patches — only cmake configuration flags are needed.
+
+## Prerequisites
+
+- **Hardware**: Any rv64gc board with 16+ GB RAM (tested on Banana Pi F3,
+  SpacemiT K1)
+- **OS**: Debian Trixie or equivalent with riscv64 packages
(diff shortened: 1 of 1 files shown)


gounthar:
Closing this docs PR as you rightly noted it is mostly redundant with the README. The real value is in CI integration, which I will focus on instead.

PR #609, opened 2026-03-11
Attempt to fix finding libedit on macOS

alexcrichton:
First pass at trying to fix this...


cmake/wasi-sdk-toolchain.cmake
```diff
@@ -113,6 +113,9 @@ if(WASI_SDK_LLDB)
     include(ProcessorCount)
     ProcessorCount(nproc)
     find_program(MAKE_EXECUTABLE make REQUIRED)
+    if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+      set(libedit_ldflags -Wl,-install_name,@rpath/libedit.0.dylib)
```
cfallin: Do we set the rpath to `../lib` somewhere (or does CMake do that automatically)? Or should this be `@rpath/../lib/libedit.0.dylib`?

PR #621, opened 2026-04-05
ci: add riscv64-linux build via cross-compilation on ubuntu-24.04

### LINKED ISSUE #607: Add riscv64 host platform support

gounthar:
Rework of the riscv64-linux CI build to use CMake cross-compilation on a standard `ubuntu-24.04` runner, rather than a native RISE runner.

## What changed

**`ci/docker/Dockerfile.riscv64-linux`** (new):
- Ubuntu 24.04 base — has `crossbuild-essential-riscv64` in its package repos
- Sets `CC=riscv64-linux-gnu-gcc` / `CXX=riscv64-linux-gnu-g++` so CMake detects cross-compilation and causes LLVM to build a native `llvm-tblgen` first, then cross-compile the rest of the toolchain
- Sets `CARGO_TARGET_RISCV64_UNKNOWN_LINUX_GNU_LINKER` for Rust cross-builds
- `XDG_CACHE_HOME=/tmp/cache` avoids write permission issues in the container

**`ci/docker-build.sh`**:
- Select `ci/docker/Dockerfile.<artifact>` if it exists, fall back to the default `ci/docker/Dockerfile`
- Make the wasmtime volume mount conditional on `WASI_SDK_CI_SKIP_SYSROOT != 1`

**`.github/workflows/main.yml`**:
- New `riscv64-linux` matrix entry: `os: ubuntu-24.04`, `rust_target: riscv64-unknown-linux-gnu`
- `cross_cmake_args: -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=riscv64 -DWASI_SDK_LLDB=OFF`
- `WASI_SDK_CI_SKIP_SYSROOT: 1`
- Handle `cross_cmake_args` in the cmake flags step

## Why WASI_SDK_CI_SKIP_SYSROOT

The cross-compiled clang runs on riscv64, not on the x86_64 build host, so the wasm sysroot step is skipped.

## Why WASI_SDK_LLDB=OFF

Avoids cross-compiling libedit and libxml2 in this first iteration; can be re-enabled as a follow-up.

Closes #607

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,16 @@ jobs:
           - artifact: arm64-linux
             os: ubuntu-22.04-arm
 
+          - artifact: riscv64-linux
+            os: ubuntu-24.04
+            rust_target: riscv64-unknown-linux-gnu
+            cross_cmake_args: >-
+              -DCMAKE_SYSTEM_NAME=Linux
+              -DCMAKE_SYSTEM_PROCESSOR=riscv64
+              -DWASI_SDK_LLDB=OFF
+            env:
+              WASI_SDK_CI_SKIP_SYSROOT: 1
+
           - artifact: arm64-macos
             os: macos-14
             rust_target: aarch64-apple-darwin
@@ -84,6 +94,9 @@ jobs:
           if [ "${{ matrix.rust_target }}" != "" ]; then
             rustup target add ${{ matrix.rust_target }}
             cmake_args="$cmake_args -DRUST_TARGET=${{ matrix.rust_target }}"
+          fi
+          if [ "${{ matrix.cross_cmake_args }}" != "" ]; then
+            cmake_args="$cmake_args ${{ matrix.cross_cmake_args }}"
           fi
           echo WASI_SDK_CI_TOOLCHAIN_CMAKE_ARGS="$cmake_args" >> $GITHUB_ENV
         shell: bash
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,9 +34,14 @@ jobs:
             cross_cmake_args: >-
               -DCMAKE_SYSTEM_NAME=Linux
               -DCMAKE_SYSTEM_PROCESSOR=riscv64
+              -DCMAKE_C_COMPILER=riscv64-linux-gnu-gcc
+              -DCMAKE_CXX_COMPILER=riscv64-linux-gnu-g++
               -DWASI_SDK_LLDB=OFF
             env:
               WASI_SDK_CI_SKIP_SYSROOT: 1
+              WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS: >-
+                -DCMAKE_SYSTEM_NAME=Linux
+                -DCMAKE_SYSTEM_PROCESSOR=riscv64
 
           - artifact: arm64-macos
             os: macos-14
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -30,7 +30,7 @@ jobs:
 
           - artifact: riscv64-linux
             os: ubuntu-24.04
-            rust_target: riscv64-unknown-linux-gnu
+            rust_target: riscv64gc-unknown-linux-gnu
             cross_cmake_args: >-
               -DCMAKE_SYSTEM_NAME=Linux
               -DCMAKE_SYSTEM_PROCESSOR=riscv64
--- a/ci/docker-build.sh
+++ b/ci/docker-build.sh
@@ -15,8 +15,13 @@ fi
 
 set -x
 
-# Build the Docker imager
-docker build --tag wasi-sdk-builder ci/docker
+# Build the Docker image. Use an artifact-specific Dockerfile if one exists
+# (e.g. ci/docker/Dockerfile.riscv64-linux), otherwise use the default.
+dockerfile=ci/docker/Dockerfile
+if [ -f "ci/docker/Dockerfile.$1" ]; then
+  dockerfile="ci/docker/Dockerfile.$1"
+fi
+docker build --tag wasi-sdk-builder --file "$dockerfile" ci/docker
 
 # Perform the build in `/src`. The current directory is mounted read-write at
 # this location as well. To ensure that container-created files are reasonable
@@ -34,9 +39,11 @@ args="$args --volume $ccache_dir:/ccache:Z --env CCACHE_DIR=/ccache"
 
 # Inherit some tools from the host into this container. This ensures that the
 # decision made on CI of what versions to use is the canonical source of truth
-# for theset ools
+# for these tools.
 args="$args --volume `rustc --print sysroot`:/rustc:ro"
-args="$args --volume $(dirname $(which wasmtime)):/wasmtime:ro"
+if [ "${WASI_SDK_CI_SKIP_SYSROOT:-}" != "1" ]; then
+  args="$args --volume $(dirname $(command -v wasmtime)):/wasmtime:ro"
+fi
 
 # Pass through some env vars that `build.sh` reads
 args="$args --env WASI_SDK_CI_TOOLCHAIN_CMAKE_ARGS"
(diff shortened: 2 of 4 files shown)


gounthar:
Reworked to use CMake cross-compilation on a standard `ubuntu-24.04` runner — no custom runner needed.

The approach: a new `ci/docker/Dockerfile.riscv64-linux` installs `crossbuild-essential-riscv64` and sets `CC=riscv64-linux-gnu-gcc` / `CXX=riscv64-linux-gnu-g++`. CMake detects cross-compilation from those and handles the two-stage build (native `llvm-tblgen` first, then cross-compiles the toolchain). `docker-build.sh` selects an artifact-specific Dockerfile when one exists. The matrix entry uses `ubuntu-24.04` with `cross_cmake_args` for `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM_PROCESSOR`.

`WASI_SDK_CI_SKIP_SYSROOT=1` and `WASI_SDK_LLDB=OFF` are kept for the same reasons as before.

.github/workflows/main.yml
```diff
@@ -28,6 +28,16 @@ jobs:
           - artifact: arm64-linux
             os: ubuntu-22.04-arm
 
+          - artifact: riscv64-linux
+            os: ubuntu-24.04
+            rust_target: riscv64-unknown-linux-gnu
```
alexcrichton: This'll want to be `riscv64gc-unknown-linux-gnu`

alexcrichton (commented):
Looks good! I suspect there'll be a number of stumbling blocks to overcome in CI, but once that's green happy to take another pass.

gounthar:
The  job was failing at the `rustup target add` step:

```
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not support target 'riscv64-unknown-linux-gnu'; did you mean 'riscv64gc-unknown-linux-gnu'?
```

Two places had the wrong triple (`riscv64-unknown-linux-gnu` instead of `riscv64gc-unknown-linux-gnu`):

- `rust_target` in `.github/workflows/main.yml`
- `CARGO_TARGET_RISCV64_UNKNOWN_LINUX_GNU_LINKER` in `ci/docker/Dockerfile.riscv64-linux` (renamed to `CARGO_TARGET_RISCV64GC_UNKNOWN_LINUX_GNU_LINKER`)

The `gc` suffix is part of the standard Linux ABI target triple and required by rustup. Fixed in the latest commit.

gounthar:
The next failure is a classic LLVM cross-compilation issue. Setting `CC`/`CXX` as Docker `ENV` vars caused LLVM's native tblgen sub-build to also compile with the riscv64 cross-compiler, producing a binary that can't run on the x86_64 build host:

```
Exec format error: llvm-min-tblgen
```

When `CMAKE_CROSSCOMPILING=TRUE`, LLVM creates a separate native sub-build for tools like `llvm-tblgen` and `llvm-min-tblgen`. The native cmake invocation does not explicitly pass `-DCMAKE_C_COMPILER`, but it does inherit environment variables, so `CC=riscv64-linux-gnu-gcc` from the Dockerfile gets picked up.

Two changes (following the same pattern as the macOS matrix entries):

1. Removed `ENV CC`/`CXX` from the Dockerfile. The cross-compiler is now passed via `-DCMAKE_C_COMPILER`/`-DCMAKE_CXX_COMPILER` cmake flags. CMake cache vars are not inherited by subprocess cmake invocations, so LLVM's native build finds the host compiler.

2. Added `WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS: -DCMAKE_SYSTEM_NAME=Linux -DCMAKE_SYSTEM_PROCESSOR=riscv64` to the matrix env. This tells LLVM it is cross-compiling and triggers the native/target build split.

.github/workflows/main.yml
gounthar: Good catch on the triple -- pushed a fix for that.

There was also a second stumbling block: setting `CC`/`CXX` as Docker `ENV` vars meant LLVM's native tblgen sub-build inherited the riscv64 cross-compiler too, which gave an `Exec format error` when the build tried to run `llvm-min-tblgen` on the x86_64 host. I moved the compilers to `-DCMAKE_C/CXX_COMPILER` cmake flags and added `WASI_SDK_CI_TOOLCHAIN_LLVM_CMAKE_ARGS` with `CMAKE_SYSTEM_NAME`/`CMAKE_SYSTEM_PROCESSOR` to trigger LLVM's native/target split -- hopefully that is the right approach, but happy to rework if not.

alexcrichton:
Thanks!

gounthar:
Thanks to both of you. @alexcrichton, your suggestion to go with cross-compilation was the better direction -- I should have looked at it from the start rather than defaulting to the native runner approach. And @tschneidereit, glad you caught the wasmtime point early, I had that wrong.

alexcrichton:
@gounthar could you test https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-33 and make sure those binaries work?

gounthar:
Tested `wasi-sdk-33.0-riscv64-linux.tar.gz` on a Banana Pi F3 (SpacemiT K1, native riscv64, kernel 6.6.99):

- `clang --version`: `22.1.0-wasi-sdk` ✓
- C → WASM compile + run via wasmtime v44.0.1:
  ```
  Hello from wasi-sdk-33 on riscv64!
  ```
  Exit code: 0 ✓
- C++ → WASM compile + run:
  ```
  Hello from C++ wasi-sdk-33 on riscv64!
  ```
  Exit code: 0 ✓

Looks good from a riscv64 native machine.

PR #632, opened 2026-05-15
Add WASI_SDK_BUILD_SHARED option

rvandermeulen:
Adds a new CMake option, defaulting to `ON` so existing builds are unaffected. When set to `OFF`, `wasi-libc` is built with `-DBUILD_SHARED=OFF` (its existing `if (NOT BUILD_SHARED)` block then short-circuits the per-target `set(SHARED ON)` for `wasm32-wasi`/`wasm32-wasip1`), and the `libcxx`/`libcxxabi`/`libunwind` invocations no longer use `${pic}` directly for their `*_ENABLE_SHARED` flags — they go through a new local `shared` variable that AND's `WASI_SDK_BUILD_SHARED` with `${pic}`.

Motivation: downstream consumers that build wasi-sdk from source as part of a larger toolchain pipeline (in our case, Firefox's build system) don't ship or link against shared sysroot artifacts. The current code path provides no way to opt out, so we've been patching locally. The existing comment in `wasi-sdk-sysroot.cmake` notes shared libraries aren't supported with LTO; this doesn't change that default, it just exposes the same toggle for non-LTO builds too.

A `static` entry is added to the `build-only-sysroot` CI matrix to exercise the OFF path.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -181,6 +181,8 @@ jobs:
           - name: default
           - name: exceptions
             defines: -DWASI_SDK_EXCEPTIONS=ON
+          - name: static
+            defines: -DWASI_SDK_BUILD_SHARED=OFF
     steps:
       - uses: actions/checkout@v6
         with:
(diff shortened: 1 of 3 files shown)


alexcrichton (approved):
Thanks for this!

PR #636, opened 2026-06-12
More work towards a wasip3 target

alexcrichton:
This is a work-in-progress to get things enabled. The main bulk of the change here is updating LLVM itself to tip-of-tree as of yesterday morning. This removes two `*.patch` files because they've landed, and adds another `*.patch` file to work around a new issue that's cropped up. I've also included a `*.patch` for https://github.com/llvm/llvm-project/pull/200855, too.

This isn't ready yet to land as it's still failing to build/link/etc, and I'll comment some more below.

PR #640, opened 2026-06-25
Experimental threading toolchain-only branch

TartanLlama:
Enables the creation of releases prefixed with `wasi-sdk-experimental-threading`, and disables building of the sysroot in CI, such that the release will contain only the compiler. This should be merged into a branch (e.g. `experimental-threading`), then we can push a tag such as `wasi-sdk-experimental-threading-1` to that branch and make a release of `wasi-sdk` that we mark as pre-release.

With that release, we should be able to test builds of `wasi-libc` that require an LLVM with libcall thread context. Note that I haven't had time to test this yet, so it's possible there are some issues with this LLVM build that would preclude that.


alexcrichton:
I've folded much of this work into https://github.com/WebAssembly/wasi-sdk/pull/636 particularly w.r.t. tags and such. In the interest of keeping in-flight things minimized I'm going to close this in favor of that, but please lemme know if I missed anything.

PR #654, opened 2026-09-01
Document how to use coop threads

alexcrichton:
This is all off-by-default right now but the distribution supports this through a separate sysroot. Document what's going on for others to discover.


dicej (approved):
Might be a good time to update https://github.com/WebAssembly/wasi-libc#building-in-pthread-support also and possibly link to this document.

